PR #5, opened 2022-01-28
Integration tests

Ardweaden:
Closes https://git.sinergise.com/team-6/openeo-platform/-/issues/41

Also fixes bug in the converter that was supposed to be fixed by https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/4 (but wasn't)  - variable names were the same as node ids, and node ids could be numbers - so we need a prefix.


tests/integration_tests/test_integration.py
```diff
@@ -0,0 +1,22 @@
...
+
+import pytest
+from pg_to_evalscript import convert_from_process_graph
+
+from tests.utils import get_process_graph_json, run_evalscript
+
+
+@pytest.mark.parametrize(
+    "pg_name,example_input,expected_output",
+    [("test_graph_1", [{"B01": 3, "B02": 3}, {"B01": 5, "B02": 1}], [4, 2])],
```
dgostencnik, line 11: So the idea is to write a bunch of similar tests for each process?

dgostencnik (approved):

.github/workflows/on_push.yml
```diff
@@ -0,0 +1,12 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Run tests
+        run: docker build -t tests . && docker run tests
```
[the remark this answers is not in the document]
Ardweaden: What did you have in mind?

tests/integration_tests/test_integration.py
Ardweaden: Well, this PR is for "integration" tests, so we test the actual evalscript produced based on the process graph, but we'll be able to test other attributes as well (e.g. if the `sampleType` is set correctly, if the number of bands get calculated correctly, if the encoding/decoding works correctly etc.). This may or may not use all the defined processes. 

After this is merged I will create issues for writing all these sorts of tests.

dgostencnik (approved):

PR #9, opened 2022-02-09
Fix issues with reduceByDimension iterating over non-existent coords.

Ardweaden:
Closes https://git.sinergise.com/team-6/openeo-platform/-/issues/46

e.g.
```javascript
d = DataCube {
    dimensions = ["x", "y"]
    data = [[1,2], [3,4]]
}

d.reduceByDimension(reducer=sum, dimension="y")
> d = DataCube {
>    dimensions = ["x"]
>    data = [3, 7]
> }
```

`_select` allows a dimension coordinate to be `null`, and returns the entire dimension.

```javascript
d = DataCube {
    dimensions = ["x", "y"]
    data = [[1,2], [3,4]]
}

d._select(d.data, [0, 0])
> 1

d._select(d.data, [0, null])
> [1,2]

d._select(d.data, [null, 1])
> [2,4]
```

`reduceByDimension` iterates over all coords while keeping the entire selected dimension.

```javascript
d.reduceByDimension(reducer=sum, dimension="y")

=> 

1. reducer(d._select(d.data, [0, null])) => 3
2. reducer(d._select(d.data, [1, null])) => 7
3. ... // (finished in our case)
```

Say there are more dimensions. For example, the shape of our datacube is `[2,4,5,3]`. We wish to reduce by the second dimension. The iteration will be `[0,null,0,0], [1,null,0,0], [1,null,1,0], [1,null,2,0] ... [1,null,4,2]`.

So we add 1 on the current dimension in each step. If the index is too big for dimension size, we move to the next dimension and add 1 there. If the dimension has `null`, we skip it. 

So before,  we did the following on the end of each iteration:

```javascript
if (coords[currInd] === null) { // If dimension number `currInd` is `null`
    currInd++ // Move to next dimension
}
coords[currInd]++ // Increase the coordinate of the `currInd` dimension
```
You can clearly see, that when going from `null` to the next dimension, we would skip over index 0 and go straight to 1. 

Not only we skipped one, but index 1 might be too large if the dimension size is 1.


src/pg_to_evalscript/javascript_datacube/DataCube.js
```diff
@@ -120,6 +120,9 @@ class DataCube {
         const labels = this.dimensions[axis].labels
 
         while (true) {
+            if (coords[currInd] === null) {
+                currInd++
+            }
             if (currInd >= shape.length) {
                 break;
             }
```
zcernigoj, line 128: Not sure if this should be solved in this PR or in a separate one, but the line 130 (2 lines below this) [`dataToReduce.labels = labels`] fails when running `convert_from_process_graph()` because `this._select(newData, coords)` returns an array, and setting labels property to an array doesn't work.

zcernigoj: Error:
```
Failed to evaluate script!
evalscript.js:145: TypeError: Cannot set property 'labels' of undefined
dataToReduce.labels = labels
^
TypeError: Cannot set property 'labels' of undefined
at DataCube.reduceByDimension (evalscript.js:145:27)
at reduce_dimension (evalscript.js:261:13)
at reduce_dimension_3aae4ee3ae1d4d5b8a5c1f38ef93f961 (evalscript.js:265:10)
at evaluatePixel (evalscript.js:270:16)
at executeForMultipleScenes (<anonymous>:1153:14)
```

Evalscript:
```javascript
//VERSION=3
function setup() {
  return {
    input: ['B01'],
    output: {
      bands: 1,
      sampleType: "FLOAT32"
    },
    mosaicking: "ORBIT"
  };
}

function updateOutput(outputs, collection) {
  Object.values(outputs).forEach((output) => {
    output.bands = 1 + 1;
  });
}
class DataCube {
  constructor(data, bands_dimension_name, temporal_dimension_name, fromSamples) {
    this.TEMPORAL = "temporal"
    this.BANDS = "bands"
    this.OTHER = "other"
    this.bands_dimension_name = bands_dimension_name;
    this.temporal_dimension_name = temporal_dimension_name;
    this.dimensions = [{
      name: this.temporal_dimension_name,
      labels: [],
      type: this.TEMPORAL
    }, {
      name: this.bands_dimension_name,
      labels: [],
      type: this.BANDS
    }]
    if (fromSamples) {
      this.data = this.makeArrayFromSamples(data)
    } else {
      this.data = data;
    }
  }

  getDimensionByName(name) {
    return this.dimensions.find(d => d.name === name)
  }

  makeArrayFromSamples(samples) {
    if (Array.isArray(samples)) {
      let newData = []
      for (let entry of samples) {
        if (this.getDimensionByName(this.bands_dimension_name).labels.length === 0) {
          this.getDimensionByName(this.bands_dimension_name).labels = Object.keys(entry)
        }
        newData.push(Object.values(entry))
      }
      return newData
    } else {
      if (this.getDimensionByName(this.bands_dimension_name).labels.length === 0) {
        this.getDimensionByName(this.bands_dimension_name).labels = Object.keys(samples)
      }
      return Object.values(samples)
    }
  }

  getBandIndices(bands) {
    return bands.map(b => this.getDimensionByName(this.bands_dimension_name).labels.indexOf(b))
  }

  filterBands(bands) {
    const indices = this.getBandIndices(bands)
    const axis = this.dimensions.findIndex(e => e.name === this.bands_dimension_name)
    this.data = this._filter(this.data, axis, indices)
    this.getDimensionByName(this.bands_dimension_name).labels = bands;
  }

  removeDimension(dimension) {
    this.dimensions = this.dimensions.filter(d => d.name !== dimension)
  }

  addDimension(name, label, type) {
    this.data = this._addDimension(this.data, 0)
    this.dimensions.unshift({
      name: name,
      labels: [label],
      type: type
    })
  }

  clone() {
    const copy = new DataCube(this._iter(this.data), this.bands_dimension_name, this.temporal_dimension_name)
    copy.dimensions = JSON.parse(JSON.stringify(this.dimensions))
    return copy
  }

  flattenToArray() {
    const flattenedData = this.flatten();
    return Array.isArray(flattenedData) ? flattenedData : [flattenedData]
  }

  encodeData() {
    const shape = this.getDataShape();
    const flattenedData = this.flattenToArray();
    return [...shape, ...flattenedData];
  }


  _iter(arr, execute = v => v, removeDim = () => false, execute_on_result = r => r, coords = []) {
    if (Array.isArray(arr)) {
      let result = []
      for (let i = 0; i < arr.length; i++) {
        const val = this._iter(arr[i], execute, removeDim, execute_on_result, [...coords, i])
        if (val !== undefined) {
          result.push(val)
        }
      }
      result = execute_on_result(result, coords)
      if (result.length === 0) {
        return
      }
      const shouldRemoveDim = removeDim(coords)
      if (shouldRemoveDim && result.length !== 1) {
        throw new Error('Can only remove dimension of length 1!')
      } else if (shouldRemoveDim) {
        return result[0]
      }
      return result
    } else {
      return execute(arr, coords)
    }
  }

  reduceByDimension(reducer, dimension) {
    let newData = this._iter(this.data)
    const axis = this.dimensions.findIndex(e => e.name === dimension)
    const shape = this.getDataShape()
    const coords = new Array(shape.length).fill(0);
    coords[axis] = null;
    let currInd = 0;
    const labels = this.dimensions[axis].labels

    while (true) {
      if (currInd >= shape.length) {
        break;
      }
      const dataToReduce = this._select(newData, coords)
      dataToReduce.labels = labels
      const newVals = reducer({
        data: dataToReduce
      })
      newData = this._set(newData, newVals, coords)
      if (coords[currInd] + 1 >= shape[currInd]) {
        currInd++
      }
      if (coords[currInd] === null) {
        currInd++
      }
      coords[currInd]++
    }
    const finalSelectCoords = new Array(shape.length).fill(null);
    finalSelectCoords[axis] = 0;
    this.data = this._select(newData, finalSelectCoords);
    this.removeDimension(dimension)
  }

  flatten() {
    const flattenArr = (arr) => arr.reduce((flat, next) => flat.concat(next), []);
    return this._iter(this.data, v => v, () => false, r => {
      return flattenArr(r)
    })
  }

  getDataShape() {
    let dimensions;
    this._iter(this.data, (a, coords) => {
      dimensions = coords;
      return a
    })
    return dimensions.map(d => d + 1)
  }

  _addDimension(arr, axis) {
    // new dimension is added before the axis
    if (!Array.isArray(arr)) {
      return [arr]
    }
    return this._iter(arr, v => v, () => false, (r, c) => {
      if (c.length === axis) {
        return [r]
      }
      return r
    })
  }

  _select(arr, coordArr) {
    // coordArr: 1D list of n coordinates. If m-th place has `null`, the entire axis is included and the dimension is kept
    function coordInSlice(c1, sliceArr) {
      return sliceArr.every((e, i) => e === null || e === c1[i])
    }
    return this._iter(arr, (a, coords) => {
      if (coordInSlice(coords, coordArr)) {
        return a
      }
    }, coords => coords.length >= coordArr.length || coordArr[coords.length] === null ? false : true)
  }

  _set(arr, vals, coordArr) {
    // Set values at coordArr
    function coordInSlice(c1, sliceArr) {
      return c1.length === sliceArr.length && sliceArr.every((e, i) => e === null || e === c1[i])
    }
    const exec_set = (a, coords) => {
      if (coordInSlice(coords, coordArr)) {
        let valueToSet;
        if (Array.isArray(vals)) {
          valueToSet = this._select(vals, coordArr.map((c, i) => c === null ? coords[i] : null).filter(c => c !== null))
        } else {
          valueToSet = vals
        }
        return valueToSet
      }
      return a
    }
    return this._iter(arr, exec_set)
  }

  _filter(arr, dim, coordArr) {
    const exec_filter = (a, coords) => {
      if (coordArr.includes(coords[dim])) {
        return a
      }
    }
    return this._iter(arr, exec_filter)
  }
}

function reduce_dimension_3aae4ee3ae1d4d5b8a5c1f38ef93f961(arguments) {

  function reduce_dimension(arguments) {
    function reducer(arguments) {

      function mean_6ff2b00fce7c4ee1ae6ca4be35d3d819(arguments) {
        function mean(arguments) {
          const {
            data
          } = arguments;
          return data.reduce((prev, curr) => prev + curr) / data.length
        }
        return mean(arguments)
      }

      let node_1 = mean_6ff2b00fce7c4ee1ae6ca4be35d3d819({
        "data": arguments.data
      })
      return node_1;
    }

    const {
      data,
      dimension
    } = arguments;
    const newData = data.clone()
    newData.reduceByDimension(reducer, dimension)
    return newData;
  }

  return reduce_dimension(arguments)
}

function evaluatePixel(samples) {
  let node_1 = new DataCube(samples, 'bands', 't', true)
  let node_2 = reduce_dimension_3aae4ee3ae1d4d5b8a5c1f38ef93f961({
    "data": node_1,
    "dimension": "t",
    "reducer": {
      "process_graph": {
        "1": {
          "process_id": "mean",
          "arguments": {
            "data": {
              "from_parameter": "data"
            }
          },
          "result": true
        }
      }
    }
  })
  return node_2.encodeData()
}
```

Ardweaden: > fails, because this._select(newData, coords) returns an array, and setting labels property to an array doesn't work.

This should work perfectly fine and this is exactly what we're doing there.

Judging from your generated evalscript, the issue is exactly the one this PR is fixing. But the evalscript was not generated with the code in this branch, but some other (older one). You can check the if statements in `reduceByDimension` of your evalscript.

zcernigoj: Thanks, not sure what happened. I was on the correct branch, but somehow the changes were not used.

I removed Pipenv.lock, removed the pipenv environment and went from the beginning and now it works as it should. I think.
The generated tiff has 2 bands, the first one has 1 for all pixels, the second one holds the actual data. Not sure what the first band is.

Ardweaden: Probably just `pipenv install` wasn't run, that would actually install the version in the branch.

> Edit: ah I see, it's the shape, the encodeData() returns [...shape, ...flattenedData] - is that ok though?

Indeed. `convert_from_process_graph` allows specifying `encode_result=False`, in which case it returns just the flattened array. That's what we do in the backend. 

Nonetheless, the output bands that are set might not be correct. It's tricky to estimate what the dimensions of the final datacube will be, so it gives the upper estimate, which means there can be extra empty bands. We'll have to handle that at some point. For example, we might actually encode the data, and then reconstruct the correct shape from it when returning results. This would then also allow to use other file formats etc. But this is down the line.

I should say that this was implemented because the original use case for this library was different - it was meant to be used as an optimization in the VITO backend. So if parts of the process graph could be converted to the evalscript, they would do so and request the data. `Evalscript` class provides `get_decoding_function` that returns a function which converts encoded process results in the correct ndarray shape, so the output could then be used as in input into the rest of the processing chain (the part that couldn't be converted).

zcernigoj: Thanks! Then this behaviour sounds correct.

zcernigoj (approved):

PR #12, opened 2022-02-11
Add details for running notebooks in the readme

zcernigoj:
I had some troubles with the commands for running the notebooks, so I added some details for alternative ways.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,3 +18,6 @@ pg-to-evalscript = {path = "."}
 
 [requires]
 python_version = "3.8"
+
+[pipenv]
+allow_prereleases = true
--- a/README.md
+++ b/README.md
@@ -184,25 +184,7 @@ convert_from_process_graph(
 ## Running notebooks
 
 ```
-pipenv install --dev
-```
- 
-Black will fail, run
-
-```
-pipenv lock --pre
-```
-
-Run setup
-
-```
-pipenv run setup.py install
-```
-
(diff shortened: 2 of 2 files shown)


Ardweaden (commented):
I think this should no longer be needed as it was added to Pipfile: https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/10f20d094fbb52f07c22edda3b75cfdf71d5e012/Pipfile#L17

So now just `pipenv install --dev` and `pipenv lock --pre` should suffice. Can you check just `pipenv install --dev --pre`?

zcernigoj:
> I think this
> 
> https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/10f20d094fbb52f07c22edda3b75cfdf71d5e012/README.md?plain=1#L197
> should no longer be needed as it was added to Pipfile:
> 
> https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/10f20d094fbb52f07c22edda3b75cfdf71d5e012/Pipfile#L17
> 
> So now just `pipenv install --dev` and `pipenv lock --pre` should suffice. Can you check just `pipenv install --dev --pre`?

Running only `pipenv install --dev --pre`, `cd notebooks` and `pipenv run jupyter notebook` works, so I updated the section, so it's shorter and cleaner

Ardweaden (approved):

PR #13, opened 2022-02-11
Add Github action to publish pg_to_evalscript on Pypi

Ardweaden:
Closes https://git.sinergise.com/team-6/openeo-platform/-/issues/50

Made release `0.1.0` https://pypi.org/project/pg-to-evalscript/0.1.0/#history

Was also tested on another branch [feature/publish-test-pypi](https://github.com/openEOPlatform/openeo-pg-evalscript-converter/tree/feature/publish-test-pypi) (`0.0.1rc5`): https://test.pypi.org/project/pg-to-evalscript/#history

Based on:

- https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
- https://github.com/sentinel-hub/sentinelhub-js/blob/5c8aeafd3af8e88f8bbe830b996b6ff9d3981442/.github/workflows/on_release.yaml

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -7,13 +7,13 @@ on:
 jobs:
   build-and-publish:
     name: Build and publish to PyPI
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Set up Python 3.8
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.8
 
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -1,0 +1,39 @@
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-and-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install pypa/build
+      run: >-
+        python -m
+        pip install
+        build
+        --user
+
+    - name: Build a binary wheel and a source tarball
+      env:
+          GITHUB_REF_TAG: ${{ github.ref }}
+      run: |
+        echo "Extracting version from GITHUB_REF_TAG: $GITHUB_REF_TAG"
+        export PG_TO_EVALSCRIPT_VERSION=$(echo "$GITHUB_REF_TAG" | sed -e "s#^refs/tags/##")
+        python -m build --sdist --wheel --outdir dist/ .
+
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,39 @@
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-and-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-18.04
```
zcernigoj, line 10: Any reason to use such an old version?
Using `ubuntu-latest` (the latest LTS version) would probably be ok.

.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    steps:
+    - uses: actions/checkout@v1
```
zcernigoj, line 13: Just a possible improvement (to be "up-to-date" longer).

We could probably use [actions/checkout@v2](https://github.com/marketplace/actions/checkout) without any change needed.

FFTR

.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
```
zcernigoj, line 16: Just a possible improvement (to be "up-to-date" longer).

We could probably use [actions/setup-python@v2](https://github.com/marketplace/actions/setup-python) without any change needed.

FFTR

zcernigoj (changes requested):
Only my first comment is important. Others are possible improvements

zcernigoj:
Just a possible improvement.

Would be nice to have a description on [pypi](https://pypi.org/project/pg-to-evalscript/0.1.0/#description). 
[Instructions for preparing a pypi friendly readme](https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/) that I found

.github/workflows/publish_to_pypi.yml
Ardweaden: I just copied it over from the linked tutorial. To be honest, I would actually prefer to have the version locked, as that should be guaranteed to always work, while a dynamic version might introduce unexpected (breaking) changes.

Ultimately, it doesn't really matter, so I changed it to `ubuntu-latest`.

We do also list 3.8 as our preferred python version in the Pipfile.

Ardweaden:
> Just a possible improvement.
> 
> Would be nice to have a description on [pypi](https://pypi.org/project/pg-to-evalscript/0.1.0/#description). Can probably be the same as the readme in this repo. [Instructions for preparing a pypi friendly readme](https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/) that I found.
> 
> FFTR.

Can be done in a separate PR.

.github/workflows/publish_to_pypi.yml
zcernigoj: > To be honest, I would actually prefer to have the version locked, as that should be guaranteed to always work, while a dynamic version might introduce unexpected (breaking) changes.

I agree. In that case I would suggest to lock it to Ubuntu 20.04 (the current LTS).
The new LTS will be 22.04, which is coming out this year and 18.04 support ends next year. But I think there would be any breaking changes related to the things we need it to do in these GitHub workflows, so `ubuntu-latest` is fine.

Resolving

zcernigoj (approved):

PR #14, opened 2022-02-11
Array_concat process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/62

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/array_concat.js
+++ b/src/pg_to_evalscript/javascript_processes/array_concat.js
@@ -10,5 +10,8 @@ function array_concat(arguments) {
       "Mandatory argument `array2` is either null or not defined."
     );
   }
+  if (!Array.isArray(array1) || !Array.isArray(array2)) {
+    throw new Error("Argument `array1` or `array2` is not an array.");
+  }
   return [...array1, ...array2];
 }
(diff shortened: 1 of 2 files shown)


dgostencnik (changes requested):
similar as in  https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/15 check mandatory inputs

zansinergise:
> similar as in #15 check mandatory inputs

Fixed the process accordingly and added additional test cases

src/pg_to_evalscript/javascript_processes/array_concat.js
```diff
@@ -0,0 +1,14 @@
+function array_concat(arguments) {
+  const { array1, array2 } = arguments;
+  if (array1 === null || array1 === undefined) {
+    throw new Error(
+      "Mandatory argument `array1` is either null or not defined."
+    );
+  }
+  if (array2 === null || array2 === undefined) {
```
dgostencnik, line 8: Perhaps it would also make sense to check if both arguments are arrays. fftr

dgostencnik (approved):

src/pg_to_evalscript/javascript_processes/array_concat.js
zansinergise: Yeah that makes sense. Added and will also add for #15

dgostencnik: :+1:

PR #15, opened 2022-02-11
Array_contains process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/63

### DIFF
--- a/tests/unit_tests/test_array_contains.py
+++ b/tests/unit_tests/test_array_contains.py
@@ -35,7 +35,8 @@ def test_array_contains(array_contains_process_code, example_input, expected_out
         ({'data': [1,2,3], 'value': 2}, False, None),
         ({'value': 2}, True, 'Mandatory argument `data` is either null or not defined.'),
         ({'data': None, 'value': 2}, True, 'Mandatory argument `data` is either null or not defined.'),
-        ({'data': [1,2,3]}, True, 'Mandatory argument `value` is not defined.')
+        ({'data': [1,2,3]}, True, 'Mandatory argument `value` is not defined.'),
+        ({'data': "[1,2,3]", 'value': 2}, True, "Argument `data` is not an array.")
     ]
 )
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_array_contains.py
```diff
@@ -0,0 +1,30 @@
...
+    return load_process_code("array_contains")
+
+
+@pytest.mark.parametrize(
+    "example_input,expected_output",
+    [
+        ({'data': [1,2,3], 'value': 2}, True),
+        ({'data': ["A","B","C"], 'value': "b"}, False),
+        ({'data': [1,2,3], 'value': "2"}, False),
+        ({'data': [1,2,None], 'value': None}, False),
```
dgostencnik, line 19: This is not in line with [Example 4](https://docs.openeo.cloud/processes/#array_contains), however it seems that either example 4 or operation definition is wrong (last bullet).

dgostencnik:
additional cases that should be handled (red asterisk marks mandatory argument) 
 - data is undefined ({"value": "2"}, raise_ex),
 - data is null ({"data": None, "value": "2"}, raise_ex),
 - value is undefined ({"data": [1, 2, 3]}, raise_ex),

tests/unit_tests/test_array_contains.py
zansinergise: I went with the assumption that the definition is correct

zansinergise:
Fixed the process accordingly and added these test cases

zansinergise:
> additional cases that should be handled (red asterisk marks mandatory argument)
> 
> * data is undefined ({"value": "2"}, raise_ex),
> * data is null ({"data": None, "value": "2"}, raise_ex),
> * value is undefined ({"data": [1, 2, 3]}, raise_ex),

Fixed the process accordingly and added these test cases

dgostencnik (approved):

PR #16, opened 2022-02-14
First and last process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/64

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/last.js
+++ b/src/pg_to_evalscript/javascript_processes/last.js
@@ -1,0 +1,17 @@
+function last(arguments) {
+  const { data, ignore_nodata = true } = arguments;
+
+  if (data === null || data === undefined) {
+    throw new Error("Mandatory argument `data` is either null or not defined.");
+  }
+
+  if (data.length === 0) {
+    return null;
+  }
+
+  if (ignore_nodata) {
+    return [...data].reverse().find((x) => x !== null);
+  }
+
+  return [...data].reverse()[0];
+}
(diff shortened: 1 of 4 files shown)


src/pg_to_evalscript/javascript_processes/last.js
```diff
@@ -0,0 +1,17 @@
...
+
+  if (data.length === 0) {
+    return null;
+  }
+
+  if (ignore_nodata) {
+    return [...data].reverse().find((x) => x !== null);
+  }
+
+  return [...data].reverse()[0];
```
dgostencnik, line 16: I think there is more efficient way to get last element of an array.

tests/unit_tests/test_first.py
```diff
@@ -0,0 +1,48 @@
...
+@pytest.fixture
+def first_process_code():
+    return load_process_code("first")
+
+
+@pytest.mark.parametrize(
+    "example_input,expected_output",
+    [
+        ({'data': []}, None),
+        ({'data': [1,2,3]}, 1),
```
dgostencnik, line 17: Can you also add test for an array will only null values

```
An array without non-null elements resolves always with null.
```

dgostencnik (changes requested):

src/pg_to_evalscript/javascript_processes/last.js
zansinergise: Should be fixed

tests/unit_tests/test_first.py
zansinergise: Should be fixed

src/pg_to_evalscript/javascript_processes/last.js
dgostencnik: Sorry, I should have been more clear about it. What I had in mind is good old 
```
data[data.length-1] 
```
fftr

zansinergise: According to google, `pop()` is supposed to be the fastest/most efficient way to get the last element of the array, so I went with this.

PR #21, opened 2022-02-16
List supported processes.

Ardweaden:
Closes https://git.sinergise.com/team-6/openeo-platform/-/issues/57

* Changes in `/unit_tests` are due to linting
* Adds instructions how to run the package in editable mode for debugging/developing to readme.


README.md
```diff
@@ -219,3 +224,10 @@ pytest
 ```
 pipenv run black -l 120 .
 ```
+
+## Developing
+
+Install the package in editable mode so the changes take effect immediately.
+```
+pipenv install -e .
```
dgostencnik, line 232: This doesn't work for me. I still have to reinstall after each change in *.py files in `src/pg_to_evalscript`.

dgostencnik (approved):

README.md
Ardweaden: This is very strange. For me it works correctly. After running `pipenv install -e .` `Pipfile` and `Pipfile.lock` now contain ` {editable = true, path = "."}` and if I change the files (and save changes), then the changes are visible when running tests.

PR #22, opened 2022-02-16
Feature/add comparison processes

dgostencnik:
https://git.sinergise.com/team-6/openeo-platform/-/issues/72

- loading "common" javascript code was added as discussed here https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/11#discussion_r804607340

-  time-strings ( 00:00:00Z) are not handled yet, but I think this use case is too obscure


src/pg_to_evalscript/javascript_processes/gt.js
```diff
@@ -0,0 +1,31 @@
...
+
+  if (x === null || y === null) {
+    return null;
+  }
+
+  if (
+    (typeof x !== "number" && typeof x !== "string") ||
+    (typeof y !== "number" && typeof y !== "string")
+  ) {
+    return false;
```
Ardweaden, line 20: Specifically

> If any operand is not a number or **temporal string** (date, time or date-time), the process returns false.

So I assume `gt({"x":"b", "y": "a"})` should return `false`.

Similar for `lt`.

src/pg_to_evalscript/javascript_processes/gte.js
```diff
@@ -0,0 +1,31 @@
...
+
+  if (x === null || y === null) {
+    return null;
+  }
+
+  if (
+    (typeof x !== "number" && typeof x !== "string") ||
+    (typeof y !== "number" && typeof y !== "string")
+  ) {
+    return false;
```
Ardweaden, line 20: Similar thing here

> If the operands are not [eq](https://docs.openeo.cloud/processes/#eq)ual (see process eq) and any of them is not a number or temporal string (date, time or date-time), the process returns false.

So I guess `gt({"x":"b", "y": "a"})` should return `False` but `gt({"x":"a", "y": "a"})` should return `true`.

Similar for `lte`.

src/pg_to_evalscript/javascript_common/common.js
```diff
@@ -0,0 +1,42 @@
+function parse_rfc3339(dt, default_h = 0, default_m = 0, default_s = 0) {
```
Ardweaden, line 1: It would be great if unit tests would be added for `parse_rfc3339` as well, given it's so often used. I trust it works correctly now, but just so we don't break anything in the future if we make any changes (e.g. for supporting time strings).

src/pg_to_evalscript/javascript_processes/between.js
```diff
@@ -0,0 +1,49 @@
...
+    return result;
+  };
+
+  const { x, min, max, exclude_max = false } = arguments;
+
+  if (x === null) {
+    return null;
+  }
+
+  if (x === undefined) {
```
Ardweaden, line 16: These checks if a required argument is defined are repeated very often, so it should perhaps be considered to move it to a function in `common`. 

(we did actually have something similar in the old python definitions of processes https://git.sinergise.com/team-6/openeo-sentinelhub-python-driver/-/blob/d6e89c3bc5a6a076406f88622533407e5bf4fea1/workers/process/eq.py#L15)

FFTR

Ardweaden (changes requested):

src/pg_to_evalscript/javascript_processes/eq.js
```diff
@@ -0,0 +1,38 @@
...
+
+  if (typeof x !== typeof y) {
+    return false;
+  }
+
+  if (
+    (typeof x !== "number" && typeof x !== "string") ||
+    (typeof y !== "number" && typeof y !== "string")
+  ) {
+    return false;
```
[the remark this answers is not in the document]
dgostencnik: booleans were added for `eq`, `neq`, `lte`, `gte`

src/pg_to_evalscript/javascript_processes/gt.js
dgostencnik: was added, but we will have to think of something else if there will be more interdependent

src/pg_to_evalscript/javascript_processes/gte.js
dgostencnik: was added , but we will have to think of something else if there will be more interdependent processes

src/pg_to_evalscript/javascript_common/common.js
dgostencnik: some tests were added.

src/pg_to_evalscript/javascript_processes/between.js
dgostencnik: I intend to do this in separate PR

src/pg_to_evalscript/javascript_processes/gte.js
Ardweaden: > but we will have to think of something else if there will be more interdependent processes

What do you mean?

tests/unit_tests/test_between.py
```diff
@@ -0,0 +1,63 @@
...
+        ({"x": None, "min": 0, "max": 1}, None),
+        ({"x": 1, "min": 0, "max": 1}, True),
+        ({"x": 1, "min": 0, "max": 1, "exclude_max": True}, False),
+        ({"x": 0.5, "min": 1, "max": 0}, False),
+        ({"x": -0.5, "min": -1, "max": 0}, True),
+        ({"x": "2018-07-23T17:22:45Z", "min": "2018-01-01T00:00:00Z", "max": "2018-12-31T23:59:59Z"}, True),
+        ({"x": "2000-01-01", "min": "2018-01-01", "max": "2020-01-01"}, False),
+        ({"x": "2018-12-31T17:22:45Z", "min": "2018-01-01T00:00:00Z", "max": "2018-12-31T23:59:59Z"}, True),
+        # Not sure why this is supposed to be True
+        # ({"x": "2018-12-31T17:22:45Z", "min": "2018-01-01", "max": "2018-12-31"}, True),
```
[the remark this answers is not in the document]
Ardweaden: Are we planning to fix this behaviour in a separate PR?

Ardweaden (commented):
Reviewing as a comment for clarification.

tests/utils.py
```diff
@@ -35,13 +35,22 @@ def run_javacript(javascript_code):
...
         return f.read()
 
 
+def load_process_code(process_id):
+    source_files = [
+        load_script("../src/pg_to_evalscript/javascript_common/", "common"),
+        load_script("../src/pg_to_evalscript/javascript_processes/", process_id),
+    ]
+
+    return reduce(lambda a, b: a + "\n" + b, source_files)
```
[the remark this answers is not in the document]
dgostencnik: was changed

tests/unit_tests/test_between.py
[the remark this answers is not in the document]
dgostencnik: should be handled now

Ardweaden (approved):

PR #23, opened 2022-02-16
Handle bands: null in load_collection

Ardweaden:
* `bands` can be `null` in `load_collection`. This means all available bands should be used. Because the converter shouldn't have any knowledge of metadata, it now raises an error if bands are `null` when `write()` is called.
* Adds a method that allows setting the `input_bands` (that also recalculates output dimensions).

Needed for https://git.sinergise.com/team-6/openeo-platform/-/issues/45


src/pg_to_evalscript/evalscript.py
```diff
@@ -119,6 +121,11 @@ def determine_output_dimensions(self):
     def set_output_dimensions(self, output_dimensions):
         self._output_dimensions = output_dimensions
 
+    def set_input_bands(self, input_bands):
```
zcernigoj, line 124: This method should probably be described in the `README.md`

Ardweaden: Indeed, added.

zcernigoj: Thanks

zcernigoj (approved):

PR #24, opened 2022-02-18
Round and int process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/80


src/pg_to_evalscript/javascript_processes/round.js
```diff
@@ -0,0 +1,26 @@
...
+
+  if (x === null) {
+    return null;
+  }
+
+  if (typeof x !== "number") {
+    throw new Error("Argument `x` is not a number.");
+  }
+
+  if (typeof p !== "number") {
```
zcernigoj, line 16: Since `p` is only allowed to be integer, the function should also check for that (`Number.isInteger(p)`) and return an error. Please also add a test for that.

zcernigoj (changes requested):
`round` needs some improvements.

src/pg_to_evalscript/javascript_processes/round.js
```diff
@@ -0,0 +1,26 @@
...
+    throw new Error("Argument `p` is not a number.");
+  }
+
+  if (p < 0) {
+    return (
+      Math.round(x / Math.pow(10, Math.abs(p))) * Math.pow(10, Math.abs(p))
+    );
+  }
+  return Number(x.toFixed(p));
```
Ardweaden, line 25: This doesn't seem to be implemented:

> **If the fractional part of x is halfway between two integers**, one of which is even and the other odd, then **the even number is returned**. This behavior follows [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229). This kind of rounding is also called "round to nearest (even)" or "banker's rounding". It minimizes rounding errors that result from consistently rounding a midpoint value in a single direction.

There is Example #4 in the documentation:
```
round(x = -2.5) => -2
```

It would be great if all the examples from the documentation were used.

Ardweaden (changes requested):

src/pg_to_evalscript/javascript_processes/round.js
zcernigoj: I don't think what you described is meant by that. Then, example 5 should be -3, not -4.

I think it's meant if the **fractional part** alternates into the infinity, like `0.12121212121212121212121...` or 0.595959595959595..., then it should do what is described with that IEEE 754 standard.

The behaviour you mentioned would be very confusing to the user.

src/pg_to_evalscript/javascript_processes/round.js
zansinergise: Added an additional check and test for this

src/pg_to_evalscript/javascript_processes/round.js
zansinergise: Added a few more tests fixed the rounding between two integers (e.g. 2.5)

src/pg_to_evalscript/javascript_processes/round.js
zcernigoj: Thanks!

Ardweaden: Is this check still needed if we have `if (!Number.isInteger(p))` check?

zansinergise: Yes I think its not needed. Removed

src/pg_to_evalscript/javascript_processes/round.js
Ardweaden: Looks fine, although I do find it very confusing that this is supposed to work only for _integers_. I would assume this sort of rounding would be done for every level, unless the floating point issues would make it infeasable.

e.g. like it's done here https://stackoverflow.com/a/3109234/5458298

Will ask for clarification at some point, but it's good for now.

Ardweaden (approved):

src/pg_to_evalscript/javascript_processes/round.js
zcernigoj: Thanks, nicely spotted, went through it too quickly. Yeah, not needed

src/pg_to_evalscript/javascript_processes/round.js
Ardweaden: Let's wait before merging, I made an issue after talking to Miha Kadunc: https://github.com/Open-EO/openeo-processes/issues/326

zcernigoj (approved):
Ah I saw I also need to approve because I also requested changes before.

Approved now

src/pg_to_evalscript/javascript_processes/round.js
Ardweaden: The issue got accepted, so I think we should implement it as in the linked stackoverflow post.

zansinergise: This should be fixed now.

Ardweaden (approved):

zcernigoj (approved):

PR #25, opened 2022-02-18
All and any process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/81


src/pg_to_evalscript/javascript_processes/all.js
```diff
@@ -0,0 +1,38 @@
...
+  if (ignore_nodata) {
+    if (data.filter((x) => x !== null).length === 0) {
+      return null;
+    }
+    return data.filter((x) => x !== null).every((x) => x === true);
+  }
+
+  if (data.length === 0) {
+    return null;
+  }
```
zcernigoj, line 27: This `if` and 

```
if (data.filter((x) => x !== null).length === 0) {
   null;
}
```
 in the `if (ignore_nodata) {` can be merged and put before the `if (ignore_nodata) {` since it should be the same behaviour regardless if `ignore_nodata` is `true` or `false`.

src/pg_to_evalscript/javascript_processes/any.js
```diff
@@ -0,0 +1,38 @@
...
+  if (ignore_nodata) {
+    if (data.filter((x) => x !== null).length === 0) {
+      return null;
+    }
+    return data.filter((x) => x !== null).some((x) => x === true);
+  }
+
+  if (data.length === 0) {
+    return null;
+  }
```
zcernigoj, line 27: Same as above - if I understand correctly, this can be optimized, so that this `if` and  `if (data.filter((x) => x !== null).length === 0) {` are before `if (ignore_nodata) {` since the behaviour is the same regardless of `ignore_nodata` value.
(cases when the array is empty or when all elements in the array are null)

zcernigoj:
A general question: So, based on the wording and the examples in the docs and already implemented processes that deal with the arrays, the assumption is that all arrays are only one-dimensional, never multi-dimensional?

Ardweaden:
@zcernigoj 

> A general question: So, based on the wording and the examples in the docs and already implemented processes that deal with the arrays, the assumption is that all arrays are only one-dimensional, never multi-dimensional?

Yes, I believe type `array` in openEO always refers to one-dimensional arrays.
These processes are typically not used directly, but are in a subprocess of a process that works on `raster-cube` type, e.g. `reduce_dimension`. `reduce_dimension` accepts `raster-cube`, which is an n-dimensional array, and a `reducer`, which is a process graph, that contains processes such as `any`/`all`/`sum` etc. The job of `reduce_dimension` is then to send one-dimensional arrays (`labeled-array` in documentation, but that's a subtype of `array`) through that `reducer` chain.

Ardweaden:
Optimisation of the code was already mentioned in this PR, and I should say that we generally didn't put too much effort into making the code fast - and there is a problem that the generated evalscripts tend to be very slow.

Even though I believe the main culprit is `_iter`, I still think it's a bit excessive that e.g. `all` would by default iterate over the same array 4 times, even though technically just once should suffice.

It's not a big deal, but I would keep these things in mind.

src/pg_to_evalscript/javascript_processes/all.js
zansinergise: Should be fixed

src/pg_to_evalscript/javascript_processes/any.js
zansinergise: Should be fixed.

zcernigoj:
> I still think it's a bit excessive that e.g. `all` would by default iterate over the same array 4 times, even though technically just once should suffice.

I think so too. The cases wouldn't look as separated as now, but something like this would lower the number of runs through the array

replacement code `all.js` after the line 11

```javascript
let returnVal = null;

for(let x of data) {
  if(!typeof x === "boolean" || x === null){
    throw new Error(...)
  }
  if(returnVal === null && x === null || returnVal === null && x === true || returnVal === true && x === null){
    returnVal = null;
  }
  if(returnVal === false || x === false){
    returnVal = false;   
  }
  if(returnVal === true || x === true){
    returnVal = true;   
  }
}

return returnVal;
```

zcernigoj (changes requested):
Sorry for requesting for changes again, but I think the benefits of having less runs through the array are big enough to make the changes (since the number of pixels grows very fast and can easily reach millions).

> I still think it's a bit excessive that e.g. `all` would by default iterate over the same array 4 times, even though technically just once should suffice.

I think so too. The cases wouldn't look as separated as now, but something like this (below) would lower the number of runs through the array

replacement code for `all.js` after the line 11

```javascript

// isn't even needed since default value of returnVal is null
// and if there is no elements in the array, the loop would do nothing 
// and returnVal would simply stay null
/*
if(data.length === 0){
  return null;
}
*/

let returnVal = null;

for(let x of data) {
  if(!(typeof x === "boolean" || x === null)){
    throw new Error(...)
  }

  // missing check for ignore_nodata and the correct behaviour based on that

  if(returnVal === null && x === null || returnVal === null && x === true || returnVal === true && x === null){
    returnVal = null;
  }
  if(returnVal === false || x === false){
    returnVal = false;   
  }
  if(returnVal === true && x === true){
    returnVal = true;   
  }
}

return returnVal;
```


I think it can be done in this MR already for `all.js` and `any.js`.

zansinergise:
> 
The code has to be a bit more complex to cover all the cases since this code misses quite a few of the edge cases. 
So I used some parts of this code and added some additional code which covers all the cases and still iterates over the array at most 1 time.

zansinergise:
> Sorry for requesting for changes again, but I think the benefits of having less runs through the array are big enough to make the changes.
> 
> > I still think it's a bit excessive that e.g. `all` would by default iterate over the same array 4 times, even though technically just once should suffice.
> 
> I think so too. The cases wouldn't look as separated as now, but something like this (below) would lower the number of runs through the array
> 
> replacement code for `all.js` after the line 11
> 
> ```js
> // isn't even needed since default value of returnVal is null
> // and if there is no elements in the array, the loop would do nothing 
> // and returnVal would simply stay null
> /*
> if(data.length === 0){
>   return null;
> }
> */
> 
> let returnVal = null;
> 
> for(let x of data) {
>   if(!(typeof x === "boolean" || x === null)){
>     throw new Error(...)
>   }
> 
>   // missing check for ignore_nodata and the correct behaviour based on that
> 
>   if(returnVal === null && x === null || returnVal === null && x === true || returnVal === true && x === null){
>     returnVal = null;
>   }
>   if(returnVal === false || x === false){
>     returnVal = false;   
>   }
>   if(returnVal === true && x === true){
>     returnVal = true;   
>   }
> }
> 
> return returnVal;
> ```
> 
> I think it can be done in this MR already for `all.js` and `any.js`.

The code has to be a bit more complex to cover all the cases since this code misses quite a few of the edge cases.
So I used some parts of this code and added some additional code which covers all the cases and still iterates over the array at most 1 time.

tests/unit_tests/test_all.py
```diff
@@ -0,0 +1,72 @@
...
+        ({"data": [False, None], "ignore_nodata": False}, False),
+        ({"data": [True, None], "ignore_nodata": False}, None),
+        ({"data": [True, False, True, False]}, False),
+        ({"data": [True, False]}, False),
+        ({"data": [True, True, True]}, True),
+        ({"data": [True]}, True),
+        ({"data": [None], "ignore_nodata": False}, None),
+        ({"data": []}, None),
+        ({"data": [], "ignore_nodata": False}, None),
+        ({"data": [True]}, True),
```
zcernigoj, line 27: Duplicate of line 23, so it can probably be deleted.

FFTR when removed.

zcernigoj (approved):
I was thinking if it would be possible to have only one `for` loop without the reducer (and deal with `ignore_nodata` in it), but it doesn't really matter since either the `for` loop or the reducer is executed. 

Good to merge in my opinion.

tests/unit_tests/test_all.py
zcernigoj: Eh, github has different workflow so every change has to be approved even if the PR was approved before. You can ping me so you won't have to wait for another approval.

zansinergise: Deleted

zansinergise:
> I was thinking if it would be possible to have only the `for` loop without the reducer (and deal with `ignore_nodata` in it), but it doesn't really matter since either the `for` loop or the reducer is executed.
> 
> Good to merge in my opinion.

Yeah I wanted to do everything in one for loop but I didn't know exactly how to do that, something always went sideways and that is why I implemented it this way because there is still only 1 iteration over the array

PR #27, opened 2022-02-22
add / SIN and COS processes

zcernigoj:
https://git.sinergise.com/team-6/openeo-platform/-/issues/84

floating point handling in javascript is ... special, so all trigonometric functions (sin, cos, tan, ...) return slightly wrong values for some of the inputs.

e.g.

math: 
`sin(PI) = 0`

js:
```javascript
> console.log(Math.sin(Math.PI));
1.2246467991473532e-16
```

Problem is most apparent for inputs `n * PI` for `n = any int value`

---

math:
`cos(PI/2) = 0`

js:
```javascript
> console.log(Math.cos(Math.PI/2));
6.123233995736766e-17
```

Problem is most apparent for inputs `(n * PI)/2` for `n = any int value`

### DIFF
--- a/tests/unit_tests/test_sin.py
+++ b/tests/unit_tests/test_sin.py
@@ -1,0 +1,49 @@
+import json
+import math
+
+import pytest
+
+from tests.utils import load_process_code, run_process
+
+
+@pytest.fixture
+def sin_process_code():
+    return load_process_code("sin")
+
+
+@pytest.mark.parametrize(
+    "example_input,expected_output",
+    [
+        ({"x": 0}, 0),
+        ({"x": math.pi / 2}, 1),
+        ({"x": math.pi}, 0),
+        ({"x": 3 * math.pi / 2}, -1),
+        ({"x": 2 * math.pi}, 0),
+        ({"x": None}, None),
+    ],
+)
+def test_sin(sin_process_code, example_input, expected_output):
+    output = run_process(sin_process_code, "sin", example_input)
+    output = json.loads(output)
+    assert output == expected_output
+
+
+@pytest.mark.parametrize(
+    "example_input,raises_exception,error_message",
+    [
+        ({"x": 0}, False, None),
+        ({}, True, "Mandatory argument `x` is not defined."),
+        ({"y": 0.5}, True, "Mandatory argument `x` is not defined."),
+        ({"x": "0.5"}, True, "Argument `x` is not a number."),
+    ],
+)
+def test_sin_exceptions(
+    sin_process_code, example_input, raises_exception, error_message
+):
+    if raises_exception:
+        with pytest.raises(Exception) as exc:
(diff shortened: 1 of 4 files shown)


Ardweaden:
Feel free to use `pytest.approx` when you stumble across issues like this.

However, while it might be that the javascript trig functions aren't returning correct values, it's also true that either python's `π` or javascript `π` aren't exactly `π`.

Of course the user won't be using either of those directly, if anything the user will use openEO's `π` (https://docs.openeo.cloud/processes/#pi) and it's up to us how we implement it. So we can do the same thing as python and javascript and just return a number with some degree of accuracy. 
Or, I was hoping we could have a special class, that we check for in the trig functions and return the value from a table. For example, by extending the `Number` class. Sadly it seems it's not possible to make custom overloads for addition/multiplication etc. operators in javascript as you can in python... 

So I don't see a good solution now. For as long we we don't specify our implementation of `π`, users can really expect to even get the exact value anyway. 

However, I don't think this inaccuracy is a particularly big issue in practice anyway.

tests/unit_tests/test_sin.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.fixture
+def sin_process_code():
+    return load_process_code("sin")
+
+
+@pytest.mark.parametrize(
+    "example_input,expected_output",
+    [
+        ({"x": 0}, 0),  # sin(0°)
+        ({"x": math.pi / 6}, pytest.approx(0.5)),  # sin(30°)
```
Ardweaden, line 18: Typically we use this directly in the assert statement, see e.g. `test_multiply` - just so you only need to write it once. But it's fine to use it like that as well, to signal explicitly that you don't expect an exact answer.

Ardweaden (approved):

zcernigoj:
> @zcernigoj It's fine, I think it's best if we don't make it too complicated for ourselves.
> 
> I was just really hoping there are was a way of doing something like this in Javascript:
> 
> ```python
> class Pi(float):
>     def __new__(self, factor=1):
>         return super().__new__(self, 3.14)
> 
>     def __init__(self, factor=1):
>         self.factor = factor
> 
>     def __add__(self, other):
>         if isinstance(other, Pi):
>             return Pi(factor=self.factor + other.factor)
>         else:
>             return super().__add__(other)
> 
>     def __mul__(self, other):
>         self.factor = self.factor * other
>         return self
> ```
> 
> But I believe it's not possible.

True, it's not possible to override or overload the existing operators in Javascript. It would be "possible" if we would make our "number" class and do all the operations with that class and at the end transform to a normal number, but that's too much work and raised complexity of everything to be worth it.

tests/unit_tests/test_sin.py
zcernigoj: Thanks, I changed this because it's easier to add tests if needed.

PR #31, opened 2022-02-22
Xor and not process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/89

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/xor.js
+++ b/src/pg_to_evalscript/javascript_processes/xor.js
@@ -21,9 +21,5 @@ function xor(arguments) {
     return null;
   }
 
-  if (x === !y) {
-    return true;
-  }
-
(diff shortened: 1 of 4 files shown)


src/pg_to_evalscript/javascript_processes/xor.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (typeof y !== "boolean" && y !== null) {
+    throw new Error("Argument `y` is not a boolean or null.");
+  }
+
+  if (x === null || y === null) {
+    return null;
+  }
+
+  if (x === !y) {
```
dgostencnik, line 24: nitpicking, this could be replaced with  
```
return x === !y;
```
fftr

dgostencnik (approved):

PR #34, opened 2022-02-24
add arctan2 process

zcernigoj:
https://git.sinergise.com/team-6/openeo-platform/-/issues/92

Reason for so many tests is that I had some difficulties finding out why passing `-0` in `test_arctan2.py` is the same as passing `0`. As mentioned in the comment in the file, python treats int `-0` same as a positive number, but float `-0.0` as a negative number.


tests/unit_tests/test_arctan2.py
```diff
@@ -0,0 +1,74 @@
...
+def arctan2_process_code():
+    return load_process_code("arctan2")
+
+
+@pytest.mark.parametrize(
+    "example_input,expected_output",
+    [
+        # test generated from examples at
+        # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2
+        # -0 has to be passed as a float so that the python handles it as a -0, not 0
```
dgostencnik, line 19: Ok, but this would make sense only if you would write your own implementation of atan2. Now you are just testing some built in function. I think it is safe to assume that Math functions working correctly and  there is no need to write so many tests for them.

dgostencnik (approved):

PR #35, opened 2022-02-24
Clip, sgn, extrema, median process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/93

### DIFF
--- a/tests/unit_tests/test_median.py
+++ b/tests/unit_tests/test_median.py
@@ -14,7 +14,13 @@ def median_process_code():
     "example_input,expected_output",
     [
         ({"data": [1, 3, 3, 6, 7, 8, 9]}, 6),
+        ({"data": [3, 7, 6, 8, 7, 1, 3]}, 6),
         ({"data": [1, 2, 3, 4, 5, 6, 8, 9]}, 4.5),
+        ({"data": [8, 6, 1, 4, 3, 9, 5, 2]}, 4.5),
+        ({"data": [1, 2, 3]}, 2),
+        ({"data": [1, 3, 2]}, 2),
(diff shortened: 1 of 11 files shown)


tests/unit_tests/test_median.py
```diff
@@ -0,0 +1,49 @@
...
+    "example_input,expected_output",
+    [
+        ({"data": [1, 3, 3, 6, 7, 8, 9]}, 6),
+        ({"data": [1, 2, 3, 4, 5, 6, 8, 9]}, 4.5),
+        ({"data": [-1, -0.5, None, 1]}, -0.5),
+        ({"data": [-1, 0, None, 1], "ignore_nodata": False}, None),
+        ({"data": []}, None),
+        ({"data": [], "ignore_nodata": False}, None),
+        ({"data": [None, None, None, None]}, None),
+        ({"data": [None, None, None, None], "ignore_nodata": False}, None),
```
dgostencnik, line 23: add following tests
```
        ({"data": [1, 2, 3]}, 2),
        ({"data": [1, 3, 2]}, 2),
        ({"data": [1, 2, 3, 4]}, 2.5),
        ({"data": [2, 1, 3, 4]}, 2.5),
```
I think we can't make an assumption that input array is sorted

dgostencnik (changes requested):

tests/unit_tests/test_median.py
zansinergise: Should be ok now, added sorting of array before calculating the median.

Added these test cases and 2 more of not sorted arrays.

dgostencnik (approved):

PR #36, opened 2022-02-24
Count process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/94

~~`Count` process still needs to be fixed, since another process needs to be applied to this one (`is_valid` and the one in `condition` variable) - need to consult on how to apply the process~~


zcernigoj:
> Count process still needs to be fixed, since another process needs to be applied to this one (is_valid and the one in condition variable) - need to consult on how to apply the process

In this case I would split this into 2 PRs, so `product` process can be merged independently.

zansinergise:
> > Count process still needs to be fixed, since another process needs to be applied to this one (is_valid and the one in condition variable) - need to consult on how to apply the process
> 
> In this case I would split this into 2 PRs, so `product` process can be merged independently.

Created a separate PR for the product process https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/40

src/pg_to_evalscript/javascript_processes/count.js
```diff
@@ -0,0 +1,44 @@
...
+  for (let val of data) {
+    if (condition === null && is_valid({ x: val })) {
+      count++;
+      continue;
+    }
+
+    if (condition === true) {
+      count++;
+      continue;
+    }
```
zcernigoj, line 35: Just a thought.

Wouldn't returning `data.length` before the `for` loop be faster?

FFTR.

zansinergise: Yes I agree, it makes sense to just return `data.length` before the for loop.

tests/utils.py
```diff
@@ -42,6 +57,13 @@ def load_script(source_file_folder, source_file_name):
         return f.read()
 
 
+def load_datacube_code():
```
zcernigoj, line 60: Is this needed already? I think it's not called anywhere.

zcernigoj (approved):

tests/utils.py
zansinergise: No it is not needed, made a mistake when I was copying another part of code. Will delete it.

zcernigoj: Thanks :+1:

Ardweaden (changes requested):
Sadly, this won't quite work. 

Both `is_valid`: 
https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/84b6ec735f80b8f24b3dcc7202bc6700b0353ede/src/pg_to_evalscript/javascript_processes/count.js#L31

and `condition` won't be defined in the generated evalscript.
https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/84b6ec735f80b8f24b3dcc7202bc6700b0353ede/src/pg_to_evalscript/javascript_processes/count.js#L36

----------------------

We have [integration tests](https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/49cd7050ed0236f68d6be60b1e86b55eab710f0a/tests/integration_tests/test_integration.py), where the evalscript is generated from process graph and then executed, to check if the correct result was produced. I suggest a new one is added.

For example, here is a (untested) process graph that uses the `count` process:
```
{
   "1":{
      "process_id":"load_collection",
      "arguments":{
         "id":"S2L2A",
         "spatial_extent":{
            "west":1,
            "south":2,
            "east":3,
            "north":4
         },
         "temporal_extent":[
            "2022-01-25T00:00:00Z",
            "2022-01-26T00:00:00Z"
         ],
         "bands":[
            "B01"
         ]
      }
   },
   "2":{
      "process_id":"reduce_dimension",
      "arguments":{
         "data":{
            "from_node":"1"
         },
         "dimension":"t",
         "reducer":{
            "process_graph":{
               "1":{
                  "process_id":"count",
                  "arguments":{
                     "data":{
                        "from_parameter":"data"
                     },
                     "condition":{
                        "gt":{
                           "arguments":{
                              "x":{
                                 "from_parameter":"element"
                              },
                              "y":2
                           },
                           "process_id":"gt",
                           "result":true
                        }
                     }
                  },
                  "result":true
               }
            }
         }
      }
   }
}
```
You can use data from Example #5 to test it.

---------------------

If you check how `pg-to-evalscript` works, you can see that we generate a `Node` class for every node in the process graph:
https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/49cd7050ed0236f68d6be60b1e86b55eab710f0a/src/pg_to_evalscript/conversion.py#L144-L178

`Node` class has a method `write`, that is called when the evalscript is being generated and which appends the source code of the process. For most processes, this just means loading the code from the definition files:
https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/49cd7050ed0236f68d6be60b1e86b55eab710f0a/src/pg_to_evalscript/node.py#L123-L131 (that comment should be removed :laughing:)

However, some processes are special, usually because they have child nodes - as they get a process graph as an argument - and need to generate the code of it as well. So they have to do a bit more in their `write` methods, load all the code for their child nodes etc. Such processes are for example `reduce_dimension` and `apply` - and now `count` of course.

E.g., here is `ReduceDimensionNode`, which inherits from `Node` but implements a couple of methods with a different behaviour:
https://github.com/openEOPlatform/openeo-pg-evalscript-converter/blob/49cd7050ed0236f68d6be60b1e86b55eab710f0a/src/pg_to_evalscript/node.py#L154-L174

tests/utils.py
```diff
@@ -73,5 +73,5 @@ def load_process_code(process_id):
 def get_defined_processes_from_files():
     return [
         os.path.splitext(os.path.basename(file_path))[0]
-        for file_path in glob.glob(f"../src/pg_to_evalscript/javascript_processes/*.js")
+        for file_path in glob.glob(f"../../src/pg_to_evalscript/javascript_processes/*.js")
```
Ardweaden, line 76: Why was this changed? Doesn't seem in scope for this PR. So that running tests from e.g. inside `integration_tests` directory works? In that case there should be a proper solution with absolute paths found. 
I assume this instruction from readme for running tests directly will no longer work:
https://github.com/openEOPlatform/openeo-pg-evalscript-converter#directly

src/pg_to_evalscript/node.py
```diff
@@ -245,3 +246,28 @@ def write_process(self):
...
+class CountNode(Node):
+    def is_process_defined(self, process_id):
+        return True
+
+    def write_process(self):
+        newline = "\n"
+        tab = "\t"
+        count_process_definition = self.load_process_code()
+        # remove first and last line from file, so we get only content of count process (function)
+        count_process_definition_content = count_process_definition[count_process_definition.find(newline) + 1 : count_process_definition.rfind('}')]
```
Ardweaden, line 260: I think this should not be necessary, you should be able to call the `count` function from inside the wrapper `count` function afaik.

Ardweaden (changes requested):

src/pg_to_evalscript/node.py
```diff
@@ -245,3 +246,28 @@ def write_process(self):
     return newData;
 }}
 """
+
+
+class CountNode(Node):
+    def is_process_defined(self, process_id):
```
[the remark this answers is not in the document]
zansinergise: Removed

tests/utils.py
zansinergise: Yeah sorry, I messed something up. Deleted this change.

Ardweaden (approved):

PR #41, opened 2022-02-28
Array_labels process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/98

### DIFF
--- a/tests/unit_tests/test_array_labels.py
+++ b/tests/unit_tests/test_array_labels.py
@@ -24,7 +24,6 @@ def test_array_labels(array_labels_process_code, example_input, expected_output)
         array_labels_process_code,
         "array_labels",
         example_input,
-        False,
         additional_js_code_to_run,
         additional_params_in_string="'data': d",
     )
@@ -50,7 +49,6 @@ def test_array_labels_exceptions(array_labels_process_code, example_input, raise
                 array_labels_process_code,
                 "array_labels",
                 example_input,
-                False,
                 additional_js_code_to_run,
                 additional_params_in_string="'data': d",
             )
@@ -61,7 +59,6 @@ def test_array_labels_exceptions(array_labels_process_code, example_input, raise
             array_labels_process_code,
             "array_labels",
             example_input,
-            False,
             additional_js_code_to_run,
             additional_params_in_string="'data': d",
         )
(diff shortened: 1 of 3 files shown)


dgostencnik:
Do you have any example for graph with labeled-array? I'm not sure if this (labels stored as data.labels)  is correct or not.

zansinergise:
> Do you have any example for graph with labeled-array? I'm not sure if this (labels stored as data.labels) is correct or not.

I do not have an example off the top of my head. I talked with Jan (a while ago) when I was implementing the [array_element](https://docs.openeo.cloud/processes/#array_element) process and he said that in the current implementation we just set the `labels` as an attribute of an array. So I guess this should be ok?

tests/unit_tests/test_array_labels.py
```diff
@@ -0,0 +1,69 @@
...
+)
+def test_array_labels_exceptions(array_labels_process_code, example_input, raises_exception, error_message):
+    additional_js_code_to_run = (
+        f"const d = {json.dumps(example_input['data'] if 'data' in example_input else 'undefined')};"
+        + f"d.labels = {json.dumps(example_input['labels']) if 'labels' in example_input else 'undefined'};"
+    )
+    process_arguments = f"{{...{json.dumps(example_input)}, 'data': d}}"
+    if raises_exception:
+        with pytest.raises(Exception) as exc:
+            run_process_with_additional_js_code(
```
dgostencnik, line 55: not sure why new function was added instead of using run_process( array_labels_process_code+ additional_js_code_to_run,...)?

zansinergise: At first this function `run_process_with_additional_js_code` was a bit different, more complex (and wrong) and it then got simplified which results in almost the same function as the original `run_process`. the main difference is that the `process_arguments` in `run_process_with_additional_js_code` is a string which contains a JS object and it can not be passed through in any other way than this, meanwhile the `example_input` in `run_process` is a python dictionary (object). So I see only 2 solutions to this problem:
- leave things as they are (not ideal)
- go through all the unit tests and replace so that `json.dumps(example_input)` (which afaik is a string) gets passed through to the `run_process` function so the `run_process_with_additional_js_code` is not needed anymore

dgostencnik: Perhaps the third option could be that run_process` would accept both, dict and string and than check `example_input` type and convert it if needed.

zansinergise: Fixed

dgostencnik (approved):

PR #42, opened 2022-03-01
Sd process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/99

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/sd.js
+++ b/src/pg_to_evalscript/javascript_processes/sd.js
@@ -36,6 +36,10 @@ function sd(arguments) {
     return null;
   }
 
+  if (count === 1) {
+    return 0;
+  }
+
   const mean = sum / count;
   let sumOfSquares = 0;
   for (let x of data) {
--- a/src/pg_to_evalscript/javascript_processes/sd.js
+++ b/src/pg_to_evalscript/javascript_processes/sd.js
@@ -1,8 +1,8 @@
 function sd(arguments) {
   const { data, ignore_nodata = true } = arguments;
 
-  if (data === undefined) {
-    throw new Error("Mandatory argument `data` is not defined.");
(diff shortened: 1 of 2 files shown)


src/pg_to_evalscript/javascript_processes/sd.js
```diff
@@ -0,0 +1,50 @@
+function sd(arguments) {
+  const { data, ignore_nodata = true } = arguments;
+
+  if (data === undefined) {
+    throw new Error("Mandatory argument `data` is not defined.");
+  }
+
+  if (!Array.isArray(data)) {
```
zcernigoj, line 8: What error should be returned when the `data` is `null`?

`"Mandatory argument `data` is not defined."` or `"Argument `data` is not an array."`

In some of the other processes the check and the error are:
```javascript
if (data === null || data === undefined) {
    throw new Error(
      "Mandatory argument `data` is either null or not defined."
    );
  }
```

src/pg_to_evalscript/javascript_processes/sd.js
```diff
@@ -0,0 +1,50 @@
...
+  let sumOfSquares = 0;
+  for (let x of data) {
+    if (ignore_nodata && x === null) {
+      continue;
+    }
+
+    sumOfSquares += Math.pow(x - mean, 2);
+  }
+
+  return Math.sqrt(sumOfSquares / (count - 1));
```
zcernigoj, line 49: Related to https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/43#discussion_r816922574

zansinergise: Added 3 additional test cases and an additional check in code to really cover all the use cases

zcernigoj (approved):

PR #44, opened 2022-03-01
add array_interpolate_linear process

zcernigoj:
https://git.sinergise.com/team-6/openeo-platform/-/issues/97

docs: https://docs.openeo.cloud/processes/?#array_interpolate_linear

A return type in the docs is `number|null` but there is nothing about propagating `null` if the array is `null` or empty (like for some of the other processes). So, I'm not sure.

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/array_interpolate_linear.js
+++ b/src/pg_to_evalscript/javascript_processes/array_interpolate_linear.js
@@ -7,6 +7,10 @@ function array_interpolate_linear(arguments) {
     );
   }
 
+  if (!Array.isArray(data)) {
+    throw new Error("Argument `data` is not an array.");
+  }
+
   const linear_interpolation = (v0, v1, t) => (1 - t) * v0 + t * v1;
   let newData = [...data];
   let start = newData[0];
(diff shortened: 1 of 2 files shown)


zcernigoj:
A return type in the docs is `number|null` but there is nothing about propagating `null` if the array is `null` (like for some of the other processes). So, I'm not sure.

dgostencnik:
I think the return type should be `array<number|null>`. For empty array I'd assume empty array is returned, for null probably null. Perhaps open an issue for that in openeo-processes repo

dgostencnik (approved):

zcernigoj:
> I think the return type should be `array<number|null>`. For empty array I'd assume empty array is returned, for null probably null. Perhaps open an issue for that in openeo-processes repo

opened an issue: https://github.com/Open-EO/openeo-processes/issues/333

zcernigoj:
> > I think the return type should be `array<number|null>`. For empty array I'd assume empty array is returned, for null probably null. Perhaps open an issue for that in openeo-processes repo
> 
> opened an issue: [Open-EO/openeo-processes#333](https://github.com/Open-EO/openeo-processes/issues/333)

Nothing to change as it already behaves as it should.

PR #45, opened 2022-03-01
Dimension_labels process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/101

### DIFF
--- a/tests/unit_tests/test_dimension_labels.py
+++ b/tests/unit_tests/test_dimension_labels.py
@@ -11,33 +11,32 @@ def dimension_labels_process_code():
 
 
 @pytest.mark.parametrize(
-    "example_input,additional_js_code_specific_to_case,expected_output",
+    "data,dimension,additional_js_code_specific_to_case,expected_output",
     [
         (
-            {"data": {"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]}, "dimension": "bands_name"},
-            "",
+            {"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]},
+            "bands_name",
+            None,
             ["B01", "B02", "B03"],
         ),
-        ({"data": {"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]}, "dimension": "temporal_name"}, "", []),
+        ({"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]}, "temporal_name", None, []),
         (
-            {
-                "data": {"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]},
-                "dimension": "temporal_name",
-            },
+            {"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]},
+            "temporal_name",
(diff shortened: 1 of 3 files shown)


tests/unit_tests/test_dimension_labels.py
```diff
@@ -0,0 +1,131 @@
...
+            {
+                "data": {"B01": [1, 2, 3], "B02": [4, 5, 6], "B03": [7, 8, 9]},
+                "dimension": "temporal_name",
+                "additional_js_code_specific_to_case": "cube.getDimensionByName('temporal_name').labels = ['test1', 'test2', 'test3'];",
+            },
+            ["test1", "test2", "test3"],
+        ),
+    ],
+)
+def test_dimension_labels(dimension_labels_process_code, example_input, expected_output):
```
Ardweaden, line 31: I find it would be more readable if not everything would be bunched together in `example_input`, but passed in separate variables `data`, `dimension` and `additional_js_code_specific_to_case`.

Ardweaden (changes requested):

tests/unit_tests/test_dimension_labels.py
zansinergise: I think the most readable way is to have `example_input` which holds all the data which should be passed to the process because if there were let's say 5 different arguments to pass to the process it could get messy if they were not grouped in an object (please do say if you disagree with me), `additional_js_code_specific_to_case` and `expected_output`

Ardweaden: Do you expect that there will be many more arguments added to this method in the future? Possibly, but I doubt it. 

Imo stuff gets cluttered because dict elements are being accessed all the time. Additionally, it's e.g. much less clear what's being passed here: `f"{{...{json.dumps(example_input)}, 'data': cube}}"` -  you have to check what the input structure is to see what the actual inputs are, it's passed as dict and then spread, and even then you realise extra parameters are being passed (`additional_js_code_specific_to_case` is not a `dimension_labels` arguments), and some arguments are being overwritten (actual `data` is `cube` and not the `data` from `example_input`).

Overall, I think the 2nd option among the following two will for example give a much clearer picture what's being done to someone who sees the code for the first time:
```python
    additional_js_code_to_run = (
        load_datacube_code()
        + f"const cube = new DataCube({example_input['data']}, 'bands_name', 'temporal_name', true);"
        + f"{example_input['additional_js_code_specific_to_case'] if 'additional_js_code_specific_to_case' in example_input else ''}"
    )
    process_arguments = f"{{...{json.dumps(example_input)}, 'data': cube}}"
```

```python
    additional_js_code_to_run = (
        load_datacube_code()
        + f"const cube = new DataCube({example_input}, 'bands_name', 'temporal_name', true);"
        + f"{additional_js_code_specific_to_case or ''}"
    )
    process_arguments = f"{{'data': cube, 'dimension': {dimension}}}"
```

I'd not a big deal for sure - but in the tests we can definitely afford to optimise for clarity as opposed to hypothetical future changes.

zansinergise: Ok yes I agree with you completely, hypothetical situations should not affect the readability of the code right now. Will fix it.

zansinergise: Fixed

Ardweaden (approved):

PR #47, opened 2022-03-03
Drop_dimension process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/106


src/pg_to_evalscript/javascript_processes/drop_dimension.js
```diff
@@ -0,0 +1,30 @@
...
+  if (typeof name !== "string") {
+    throw new Error("Argument `name` is not a string.");
+  }
+
+  const dim = data.getDimensionByName(name);
+  if (!dim) {
+    throw new Error("A dimension with the specified name does not exist.");
+  }
+
+  if (dim.labels.length > 1) {
```
Ardweaden, line 21: This should be ok in theory, but in practice we don't have labels set most of the time (we do for the bands dimension, but mostly not for others). So perhaps this should be handled better in the `DataCube`, but I think just checking the actual size of the dimension is a good enough solution.

src/pg_to_evalscript/javascript_processes/drop_dimension.js
```diff
@@ -0,0 +1,30 @@
...
+    throw new Error(
+      "The number of dimension labels exceeds one, which requires a reducer."
+    );
+  }
+
+  let newData = data.clone();
+  newData.removeDimension(name);
```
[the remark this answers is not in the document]
zansinergise: I think this should be fixed now

src/pg_to_evalscript/javascript_processes/drop_dimension.js
zansinergise: Fixed

src/pg_to_evalscript/javascript_datacube/DataCube.js
```diff
@@ -58,7 +58,32 @@ class DataCube {
...
+                for (let i = 0; i < tmpData.length; i++) {
+                    data[i] = tmpData[i]
+                }
+                return;
+            }
+
+            for (let j = 0; j < data.length; j++) {
+                const tmpData = data[j]
+                for (let i = 0; i < tmpData.length; i++) {
+                    data[j] = tmpData[i]
```
Ardweaden, line 79: For simpler understanding, everything inside the outer `for` could be rewritten as:
```js
data[j] = lastElement(data[j]) // or data[j] = data[j][data[j].length-1] using JS syntax
```
`data` needs to be an array of length 1 if we want to drop the dimension, other wise that doesn't make any sense. So it could be just `data[j] = data[j][0]`.

However, this is actually dropping the dimension a level lower than we are currently at. That's why `idx === 0 ? idx : idx - 1` is done. This of course introduces bugs.

For example, say our data is
```
[[1],[2],[3]]
```
And we want to drop the second dimension (index 1). We would expect `[1,2,3]`. However:
* calling `removeDimensionFromData(data,0,1)` doesn't do anything, as we enter this second loop and get e.g `[1]` as `data` and just sets `data[0] = 1`
* calling `removeDimensionFromData(data,0,0)` as would come from `removeDimension` enters `if (currLevel === 0)`. That one just sets the first element to `1`.

---------------------------------

Anyway, this function could also be implemented using existing methods defined on the `DataCube`, namely `_iter`.

e.g I believe this should work:
```js
removeDimension(dimension) {
    const idx = this.dimensions.findIndex(d => d.name === dimension);
    this.dimensions = this.dimensions.filter(d => d.name !== dimension);
    this.data = this._iter(this.data, v => v, coords => coords.length === idx)
}
```
It ultimately doesn't matter if it's implemented separately or not, because the implementation of `DataCube` based on arrays-of-arrays will be changed soon as a part of performance optimisations.

src/pg_to_evalscript/javascript_datacube/DataCube.js
```diff
@@ -58,7 +58,32 @@ class DataCube {
       }
 
     removeDimension(dimension) {
```
Ardweaden, line 60: `removeDimension` is also called from inside `reduceByDimension` process and there it only expects to filter out the dimension and not actually drop it. It's a testament to the fact we don't have any proper tests for `reduce_dimension`...

tests/unit_tests/test_drop_dimension.py
```diff
@@ -0,0 +1,166 @@
...
+@pytest.fixture
+def drop_dimension_process_code():
+    return load_process_code("drop_dimension")
+
+
+@pytest.mark.parametrize(
+    "data,name,additional_js_code_specific_to_case,expected_output",
+    [
+        (
+            {"B01": [1, 2, 3]},
```
Ardweaden, line 18: Btw can we change all these `data` values as they are not really valid. I noticed this got introduced in some other PRs. Namely, band values will never be arrays, but numbers. Additionally, the case where we don't have temporal script also isn't properly implemented, so we can expect the input `samples` will be arrays of objects.

Like this:
`{"B01": [1, 2, 3]}` => `[{"B01": 1}]`

-------------------------------------

Btw, the last argument in `DataCube` constructor is a bool `fromSamples`, which says if the input data should be converted from the above format to arrays of arrays or should it set it directly. If `false`, you can set it directly.
e.g.
```js
new Datacube([[[[1,2,3,4]], [[5,6,7,8]]], [[[9,10,11,12]], [[13,14,15,16]]]], 't', 'b', false)
```
However, I don't recommend doing that in these tests as we soon won't be using arrays of arrays anymore. Just generally so you're aware.

Ardweaden (changes requested):

src/pg_to_evalscript/javascript_processes/drop_dimension.js
```diff
@@ -0,0 +1,44 @@
...
+
+  function getDataForDimension(data, currLevel, dimLevel) {
+    if (currLevel === dimLevel) {
+      return data;
+    }
+    return getDataForDimension(data[0], currLevel + 1, dimLevel);
+  }
+
+  const dimLevel = data.dimensions.findIndex((d) => d.name === name);
+  const dimData = getDataForDimension(data.data, 0, dimLevel);
```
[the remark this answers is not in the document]
zansinergise: Ah I see it now, fixed

src/pg_to_evalscript/javascript_datacube/DataCube.js
zansinergise: Ah I see the bug now yes..

This solution with the `_iter` function works like a charm I think.

tests/unit_tests/test_drop_dimension.py
zansinergise: Ok I will fix these incorrect `data` inputs.

Thanks for the info!

src/pg_to_evalscript/javascript_datacube/DataCube.js
zansinergise: Should I try to fix this, or did you just mention it so I would know this?

zansinergise: I changed the `reduceByDimension` function so it now only filters the dimensions instead of actually removing it

Ardweaden (changes requested):

tests/unit_tests/test_drop_dimension.py
zansinergise: Ah I see, sorry I misunderstood. Reverted that commit so the changes are only for this file.

src/pg_to_evalscript/javascript_processes/drop_dimension.js
```diff
@@ -0,0 +1,37 @@
...
+    parameterName: "name",
+    value: name,
+    nullable: false,
+    required: true,
+    allowedTypes: ["string"],
+  });
+
+  const dim = data.getDimensionByName(name);
+  if (!dim) {
+    throw new Error("A dimension with the specified name does not exist.");
```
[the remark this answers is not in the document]
zansinergise: Changed

tests/integration_tests/test_integration.py
```diff
@@ -30,8 +30,16 @@
             [{"B01": 0.5, "B02": 0.75}, {"B01": 0, "B02": 1}, {"B01": -3, "B02": 4}, {"B01": None, "B02": None}],
             [1, 1.5, 0, 2, 0, 2, None, None],
         ),
-        ("test_count_with_condition", [{"B01": 0, "B02":1}, {"B01":2, "B02":3}, {"B01":4, "B02":5}, {"B01":None, "B02": None}], [1, 2]),
-        ("test_count_without_condition", [{"B01": 0, "B02":1}, {"B01":2, "B02":3}, {"B01":4, "B02":5}, {"B01":None, "B02": 3}], [3, 4]),
+        (
```
Ardweaden, line 33: Why was this changed?

Ardweaden (changes requested):

tests/integration_tests/test_integration.py
zansinergise: With the changes to `removeDimension` and consequently to `reduceByDimension` the latter function did not work correctly (I did not know that at the time and you pointed it out). In the meantime while I was unaware that it was not working correctly and I thought everything was ok, I changed the tests. Will change them back!

Ardweaden: Well that's why the tests are there 😄 If they fail, there should be something wrong with the code... Of course some time tests are wrong, but the first thought should be the code, especially if the output and input and the test setup are correct.

Ardweaden (approved):

PR #51, opened 2022-03-10
Add order process

mitjak:
https://git.sinergise.com/team-6/openeo-platform/-/issues/111

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/order.js
+++ b/src/pg_to_evalscript/javascript_processes/order.js
@@ -1,0 +1,37 @@
+function order(arguments) {
+  const { data, asc = true, nodata = null} = arguments;
+
+  if (data === null || data === undefined) {
+    throw new Error("Mandatory argument `data` is either null or not defined.");
+  }
+
+  if (!Array.isArray(data)) {
+    throw new Error("Argument `data` is not an array.");
+  }
+
+  if (typeof asc !== "boolean") {
+    throw new Error("Argument `asc` is not a boolean.");
+  }
+
+  if (typeof nodata !== "boolean" && nodata !== null) {
+    throw new Error("Argument `asc` is not a boolean or null.");
+  }
+
+  return data
+    .map((value, index) => {
+      return { value, index };
+    })
+    .sort((a, b) => {
+      if (a.value === null && b.value === null) return 0;
+      if (a.value === null) return nodata ? 1 : -1;
+      if (b.value === null) return nodata ? -1 : 1;
+
+      if (a.value === b.value) return 0;
+      if (a.value < b.value) return asc ? -1 : 1;
+      if (a.value > b.value) return asc ? 1 : -1;
+    })
+    .filter(obj => {
+      return obj.value !== null || nodata !== null;
+    })
+    .map(obj => obj.index);
+}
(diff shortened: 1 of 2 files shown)


src/pg_to_evalscript/javascript_processes/order.js
```diff
@@ -0,0 +1,46 @@
+function order(arguments) {
+  const { data, asc = true, nodata = null} = arguments;
+
+  if (data === null || data === undefined) {
```
zansinergise, line 4: The parameters should probably be checked with the "new" `validateParameter` function.

zansinergise (changes requested):

src/pg_to_evalscript/javascript_processes/order.js
```diff
@@ -0,0 +1,63 @@
+function order(arguments) {
+  const { data, asc = true, nodata = null} = arguments;
+
+  validateParameter({
+    processName: "order",
+    parameterName: "data",
+    value: data,
+    required: true,
```
zansinergise, line 8: `nullable: false` should be added here as well.

src/pg_to_evalscript/javascript_processes/order.js
```diff
@@ -0,0 +1,63 @@
...
+    array: true,
+  });
+
+  validateParameter({
+    processName: "order",
+    parameterName: "asc",
+    value: asc,
+    required: true,
```
zansinergise, line 16: This `required: true` might be misleading since the parameter is not actually required

src/pg_to_evalscript/javascript_processes/order.js
```diff
@@ -0,0 +1,63 @@
...
+    nullable: false,
+    allowedTypes: ["boolean"],
+  });
+
+  validateParameter({
+    processName: "nodata",
```
zansinergise, line 22: wrong process name

src/pg_to_evalscript/javascript_processes/order.js
```diff
@@ -0,0 +1,63 @@
...
+    parameterName: "nodata",
+    value: nodata,
+    required: true,
+    allowedTypes: ["boolean"],
+  });
+
+  const dataWithIndex = [];
+  for (let i = 0; i < data.length; i++) {
```
zansinergise, line 30: Checking each value in the array with the `validateParameter` function would also be quite nice IMO

zansinergise (changes requested):

src/pg_to_evalscript/javascript_processes/order.js
mitjak: Hm, this was still functioning as expected because default value is set. Nevertheless, it should be updated as I agree it's misleading.

src/pg_to_evalscript/javascript_processes/order.js
mitjak: We are already checking each value and throwing an error if there is no match.

While it does not produce the same error type I find it is more (future) error proof as value checking is done in one place instead of duplicating the condition logic.

zansinergise: Ok yes you have a point, I agree with you. Resolving

src/pg_to_evalscript/javascript_processes/order.js
zansinergise: I think it looks good to me now.

zansinergise (approved):

PR #53, opened 2022-03-11
Feature/use parameter validation

dgostencnik:


src/pg_to_evalscript/javascript_common/common.js
```diff
@@ -128,7 +128,12 @@ function validateParameter(arguments) {
     });
   }
 
-  if (boolean && typeof value !== "boolean") {
+  if (
```
mitjak, line 131: Why don't we rather use the `allowedTypes` prop for `boolean` as well? The only difference I see is different error name and that can be modified as well. As it stands we can now validate `boolean` in two different ways.

 *  *  *  *  *

On that note - `validateParameter()` interface would be much cleaner if we would use `allowedTypes` for everything and then filter internally based on custom (`array` and `integer`) or native type (all others). But it would complicate things quite a lot and not sure if it's worth it at this point.

dgostencnik: `boolean` was replaced with `allowedTypes`, not sure why I used separate prop in a first place.

mitjak: Cool. `boolean` could also be removed from arguments list in the `validateParameter` function. fftr

mitjak (approved):

PR #54, opened 2022-03-14
Array_apply process

zansinergise:
https://git.sinergise.com/team-6/openeo-platform/-/issues/114

### DIFF
--- a/src/pg_to_evalscript/node.py
+++ b/src/pg_to_evalscript/node.py
@@ -284,6 +284,16 @@ function array_apply(arguments) {{
     }}
 
     const {{data, context = null}} = arguments;
+
+    validateParameter({{
+        processName: "array_apply",
+        parameterName: "data",
+        value: data,
+        required: true,
+        nullable: false,
+        array: true
+    }});
+
     let newData = [];
     for (let i = 0; i < data.length; i++) {{
         newData[i] = process({{
--- a/src/pg_to_evalscript/node.py
+++ b/src/pg_to_evalscript/node.py
@@ -281,6 +281,7 @@ function array_apply(arguments) {{
     }}
 
     {self.load_process_code()}
-    return array_apply(arguments)
+    const {{ data, context }} = arguments;
+    return array_apply({{ data: data, context: context, process: process }})
 }}
 """
--- a/src/pg_to_evalscript/node.py
+++ b/src/pg_to_evalscript/node.py
@@ -269,9 +269,6 @@ function count(arguments) {{
 """
 
 class ArrayApplyNode(Node):
-    def is_process_defined(self, process_id):
-        return True
-        
     def write_process(self):
         newline = "\n"
         tab = "\t"
--- a/src/pg_to_evalscript/node.py
+++ b/src/pg_to_evalscript/node.py
@@ -73,6 +73,7 @@ class Node:
             "apply": ApplyNode,
             "count": CountNode,
             "array_apply": ArrayApplyNode,
+            "array_filter": ArrayFilterNode,
         }
         if class_types_for_process.get(process_id):
             self.__class__ = class_types_for_process[process_id]
@@ -269,7 +270,7 @@ function count(arguments) {{
 """
 
 class ArrayApplyNode(Node):
-    def write_process(self):
+    def write_process(self):   
         newline = "\n"
         tab = "\t"
         return f"""
@@ -283,5 +284,22 @@ function array_apply(arguments) {{
     {self.load_process_code()}
     const {{ data, context }} = arguments;
     return array_apply({{ data: data, context: context, process: process }})
+}}
+"""
+
+class ArrayFilterNode(Node):
+    def write_process(self):
+        newline = "\n"
+        tab = "\t"
+        return f"""
+function array_filter(arguments) {{
+    function condition(arguments) {{
+    {newline.join(node.write_function() for node in self.child_nodes)}
+    {newline.join(node.write_call() for node in self.child_nodes)}
+        return {self.child_nodes[-1].node_varname_prefix + self.child_nodes[-1].node_id};
+    }}
+
+    {self.load_process_code()}
+    return array_filter(arguments);
 }}
 """
--- a/src/pg_to_evalscript/node.py
+++ b/src/pg_to_evalscript/node.py
@@ -283,27 +283,7 @@ function array_apply(arguments) {{
         return {self.child_nodes[-1].node_varname_prefix + self.child_nodes[-1].node_id};
     }}
 
-    const {{data, context = null}} = arguments;
-
-    validateParameter({{
-        processName: "array_apply",
-        parameterName: "data",
-        value: data,
-        required: true,
-        nullable: false,
-        array: true
-    }});
-
-    let newData = [];
-    for (let i = 0; i < data.length; i++) {{
-        newData[i] = process({{
-            x: data[i],
-            index: i,
-            label: data.labels ? data.labels[i] : undefined,
-            context: context,
-        }});
-    }}
-
-    return newData;
+    {self.load_process_code()}
+    return array_apply(arguments)
 }}
 """
--- a/src/pg_to_evalscript/node.py
+++ b/src/pg_to_evalscript/node.py
@@ -72,6 +72,7 @@ class Node:
             "if": IfNode,
             "apply": ApplyNode,
             "count": CountNode,
+            "array_apply": ArrayApplyNode,
         }
         if class_types_for_process.get(process_id):
             self.__class__ = class_types_for_process[process_id]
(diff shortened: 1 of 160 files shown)


src/pg_to_evalscript/utils.py
```diff
@@ -3,7 +3,7 @@
 
 def list_supported_processes():
     process_definitions_directory = "javascript_processes"
-    implicitly_supported_processes = ["load_collection", "save_result", "reduce_dimension", "apply"]
+    implicitly_supported_processes = ["load_collection", "save_result", "reduce_dimension", "apply", "array_apply"]
```
zcernigoj, line 6: Not connected to this MR, but to https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/36

I think we forgot to add the `count` process.

FFTR.

zcernigoj (approved):

src/pg_to_evalscript/node.py
```diff
@@ -266,3 +267,33 @@ def write_process(self):
...
+        tab = "\t"
+        return f"""
+function array_apply(arguments) {{
+    function process(arguments) {{
+    {newline.join(node.write_function() for node in self.child_nodes)}
+    {newline.join(node.write_call() for node in self.child_nodes)}
+        return {self.child_nodes[-1].node_varname_prefix + self.child_nodes[-1].node_id};
+    }}
+
+    const {{data, context = null}} = arguments;
```
zcernigoj, line 286: I think we should check if `data` is an array.

zcernigoj (changes requested):

src/pg_to_evalscript/utils.py
zansinergise: I think this should **not** be added, since the definition of the `count` process is in the file and thus gets read from there when running this test.

zcernigoj: Ah thanks, true. Got carried away when I was checking `node.py`. Resolving

src/pg_to_evalscript/node.py
zcernigoj: The same long error stack is printed for unit tests if you put a test for errors in the method that checks valid inputs.
Tests for errors should be added (similar to the ones in unit tests), so that the error is checked.

These tests should be added:
- `data` is not array
- `data` is empty array (`[]`) - currently returns `[null]` but should return empty array (`[]`) as per docs
  > Return Value
  >
  > An array with the newly computed values. The number of elements are the same as for the original array.
- element in `data` is not appropriate for the applied process (e.g. string and `add` process)


Maybe something like this for the errors that the applied process could ~~return~~ throw:

```
    for (let i = 0; i < data.length; i++) {{
        try{{
          newData[i] = process({{
              x: data[i],
              index: i,
              label: data.labels ? data.labels[i] : undefined,
              context: context,
          }});
        }}
        catch (error_from_the_applied_process) {{
          // propagate that same error
          // or throw a "wrapper" error adding the info 
          // for which parent process and which element
          // the error in the applied process happened
        }}
    }}
```

src/pg_to_evalscript/utils.py
Ardweaden: This is correct, however the way `reduce_dimension` and `apply` process are implemented isn't great and will probably be moved to separate files and implemented in a similar manner like `count`. It would be great if this was done here as well, because we can then add unit tests as well.

Ardweaden (changes requested):
As mentioned in https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/54#discussion_r828978555

src/pg_to_evalscript/utils.py
zansinergise: Makes sense yes, will move it to a file and define some unit tests

src/pg_to_evalscript/node.py
zansinergise: Added all these additional test cases and a few more. Can you explain please how you got to the result of the second test case that you mentioned, where data is an empty array `[]`. I can't understand how this could happen, since the for loop runs "through" the data array and if it is empty, nothing should happen and an empty array should be returned.

Regarding the long stack trace, nothing changes whether I simply catch and propagate the same error or create a new error or actually just catch the error and don't do anything with it. In all cases the same long stack trace remains.

src/pg_to_evalscript/node.py
```diff
@@ -266,3 +267,23 @@ def write_process(self):
     return count(arguments);
 }}
 """
+
+class ArrayApplyNode(Node):
+    def is_process_defined(self, process_id):
```
Ardweaden, line 272: This can be removed now.

tests/process_graphs/test_array_apply_add.json
```diff
@@ -0,0 +1,49 @@
...
+              },
+              "process": {
+                "process_graph": {
+                  "add1": {
+                    "process_id": "add",
+                    "arguments": {
+                      "x": {
+                        "from_parameter": "x"
+                      },
+                      "index": {
```
Ardweaden, line 34: `add` does not accept an argument `index`, so this can be removed.

tests/unit_tests/test_array_apply.py
```diff
@@ -0,0 +1,36 @@
...
+
+@pytest.fixture
+def array_apply_process_code():
+    return load_process_code("array_apply")
+
+
+@pytest.mark.parametrize(
+    "example_input,expected_output", 
+    [
+        ({"data": [], "process": {"process_graph": {"add_pg": {"process_id": "add", "arguments": {"x": {"from_parameter": "x"}, "index": {"from_parameter": "index"}, "y": 10}}}}}, [])
```
Ardweaden, line 16: I don't think this test does anything. When running unit tests the JS code is run directly, it doesn't go through the converter that would generate a JS function from the `process` argument. 
Thus, the `process` argument should be passed as a JS function. Similar to how it's done here: https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/57/files

Ardweaden (changes requested):

src/pg_to_evalscript/node.py
zansinergise: Done

tests/process_graphs/test_array_apply_add.json
zansinergise: done

tests/unit_tests/test_array_apply.py
zansinergise: This one in particular does not do much yes, it just propagates the empty array. However the `array_apply` process function checks for both parameters `data` and `process` which are required and are array/object. That is why there is an actual process_graph object as the parameter instead of some dummy text or something like that.

I added a few more tests similar to how you proposed.

mitjak: I'd say that since `process` parameter is not used it would be clearer if some dummy text would be passed. Passing a process graph that is not parsed or used might be misleading.

On that note, I am not entirely sure that we've taken the best approach by validating the `process` parameter (the same as in the `array_filter`). We might rather (or possibly also) check if `process()` function is defined and act upon that.

Ardweaden: > I'd say that since process parameter is not used it would be clearer if some dummy text would be passed. Passing a process graph that is not parsed or used might be misleading.
On that note, I am not entirely sure that we've taken the best approach by validating the process parameter (the same as in the array_filter). We might rather (or possibly also) check if process() function is defined and act upon that.

What we could also do is change `write_process` like this:
```python
def write_process(self):
        ...
        return f"""
function array_apply(arguments) {{
    function process(arguments) {{
        ... 
    }}
    ...
    const { data, context } = arguments;
    return array_apply({ data: data, process: process, context: context })
}}
"""
```
And then we can pass the `process` as a function directly in unit tests. We can then also check if `process` is defined and callable.

mitjak: This would obviously be ideal. Let's go with this.

zansinergise: Should be fixed now.

tests/unit_tests/test_array_apply.py
```diff
@@ -0,0 +1,88 @@
...
+            {"data": "[1,2,3,4,5]", "process": f"({{x}}) => x + 10;"},
+            True,
+            "NOT_ARRAY",
+        ),
+        ({"data": [1, 2, 3, 4, 5], "process": 23}, True, "WRONG_TYPE"),
+    ],
+)
+def test_array_apply_inputs(array_apply_process_code, example_input, raises_exception, error_name):
+    process_js_code = (
+        f"const process = eval({json.dumps(example_input['process']) if 'process' in example_input else 'undefined'});"
```
Ardweaden, line 83: Is there a reason why `eval` was used? The `process` script could be passed directly into `process_arguments` as far as I can see.

Ardweaden (changes requested):

tests/unit_tests/test_array_apply.py
zansinergise: Yeah I thought so also, but for some reason I had to use eval before to make it work, and now when I tried to delete the eval, it works without it. I have no idea what happened

Ardweaden: You don't need `json.dumps` because it's a normal string and not an object or array. 
Also there is no need to use [formatted string literal](https://docs.python.org/3/tutorial/inputoutput.html#fancier-output-formatting) for the `process` parameter, because you aren't putting any variables inside the string and it makes it harder to read. 
`f"({{x}}) => x + 10;"` => `"({x}) => x + 10;"`

zansinergise: `json.dumps` is needed by those test cases where `data`/`process` is `None` or a `True` or something like that, so it gets formatted correctly. In cases where `process` is a string indeed it is not needed.

I agree yes, changed to normal string.

Ardweaden: Ah yes true.

Ardweaden (approved):

src/pg_to_evalscript/node.py
zcernigoj: > Can you explain please how you got to the result of the second test case that you mentioned, where data is an empty array []. I can't understand how this could happen, since the for loop runs "through" the data array and if it is empty, nothing should happen and an empty array should be returned.

I tried to get to the bottom of this but I couldn't find where the `null` (or `None` as python interprets it) comes from, so I think we can leave that to be solved later.

> Regarding the long stack trace, nothing changes whether I simply catch and propagate the same error or create a new error or actually just catch the error and don't do anything with it. In all cases the same long stack trace remains.

I think the long stack trace is correct so I think it's just the matter of catching (or not) the error in this function, which is more to help the developers / users of this project than anything else.

zcernigoj (approved):

PR #57, opened 2022-03-15
Add array filter process

mitjak:
https://git.sinergise.com/team-6/openeo-platform/-/issues/115

### DIFF
--- a/tests/integration_tests/test_integration.py
+++ b/tests/integration_tests/test_integration.py
@@ -32,6 +32,7 @@ from tests.utils import (
         ),
         ("test_count_with_condition", [{"B01": 0, "B02":1}, {"B01":2, "B02":3}, {"B01":4, "B02":5}, {"B01":None, "B02": None}], [1, 2]),
         ("test_count_without_condition", [{"B01": 0, "B02":1}, {"B01":2, "B02":3}, {"B01":4, "B02":5}, {"B01":None, "B02": 3}], [3, 4]),
+        # todo: test_array_filter process graph uses reduce_dimension which gives unconclusive results, update with apply_dimension once implemented
         ("test_array_filter", [{"B01": 0}, {"B01": 1}, {"B01":2}, {"B01":4}, {"B01":5}], [2]),
     ],
 )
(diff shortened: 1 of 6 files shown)


tests/integration_tests/test_integration.py
```diff
@@ -32,6 +32,8 @@
         ),
         ("test_count_with_condition", [{"B01": 0, "B02":1}, {"B01":2, "B02":3}, {"B01":4, "B02":5}, {"B01":None, "B02": None}], [1, 2]),
         ("test_count_without_condition", [{"B01": 0, "B02":1}, {"B01":2, "B02":3}, {"B01":4, "B02":5}, {"B01":None, "B02": 3}], [3, 4]),
+        # todo: test_array_filter process graph uses reduce_dimension which gives unconclusive results, update with apply_dimension once implemented
+        ("test_array_filter", [{"B01": 0}, {"B01": 1}, {"B01":2}, {"B01":4}, {"B01":5}], [2]),
```
Ardweaden, line 36: Perhaps we could anyway change the process graph so that it would produce a "correct" result. Because of course the output is now non-sensical as `reduce_dimension` doesn't return a single value, but an array. I was thinking something like this:
```json
{
   "loadcollection":{
      "process_id":"load_collection",
      "arguments":{
         "spatial_extent":null,
         "id":"S2L2A",
         "temporal_extent":null,
         "bands":[
            "B01"
         ]
      }
   },
   "reduce":{
      "process_id":"reduce_dimension",
      "arguments":{
         "data":{
            "from_node":"loadcollection"
         },
         "dimension":"t",
         "reducer":{
            "process_graph":{
               "array_filter":{
                  "process_id":"array_filter",
                  "arguments":{
                     "data":{
                        "from_parameter":"data"
                     },
                     "condition":{
                        "process_graph":{
                           "gt":{
                              "arguments":{
                                 "x":{
                                    "from_parameter":"x"
                                 },
                                 "y":1
                              },
                              "process_id":"gt"
                           }
                        }
                     }
                  },
                  "result":true
               },
               "sum":{
                  "process_id":"sum",
                  "arguments":{
                     "data":{
                        "from_node":"array_filter"
                     }
                  },
                  "result":true
               }
            }
         }
      }
   }
}
```
so just summing the filtered array to get a result that makes sense. Admittedly the test isn't so `array_filter` specific anymore, but correct behaviour can still be deducted from the result and these are integration tests anyway.

Ardweaden (changes requested):

tests/integration_tests/test_integration.py
mitjak: Cool, I think this makes total sense for integration tests. We have more fine grained tests for `array_filter` in unit tests already.

Ardweaden (approved):

PR #60, opened 2022-03-22
Refactor reduce dimension and apply

dgostencnik:
https://git.sinergise.com/team-6/openeo-platform/-/issues/118


src/pg_to_evalscript/javascript_processes/reduce_dimension.js
```diff
@@ -0,0 +1,31 @@
...
+
+  validateParameter({
+    processName: "reduce_dimension",
+    parameterName: "data",
+    value: data,
+    nullable: false,
+    required: true,
+  });
+
+  validateParameter({
```
mitjak, line 12: Dimension can also be validated against type as it should be a `string`.

src/pg_to_evalscript/javascript_processes/reduce_dimension.js
```diff
@@ -0,0 +1,31 @@
+function reduce_dimension(arguments) {
+  const { data, dimension, reducer } = arguments;
```
mitjak, line 2: Argument `context` could be passed to `reducer` function, same as above.

mitjak (changes requested):

mitjak (changes requested):

src/pg_to_evalscript/javascript_processes/reduce_dimension.js
dgostencnik: was added

src/pg_to_evalscript/javascript_processes/apply.js
```diff
@@ -0,0 +1,23 @@
+function apply(arguments) {
+  const { data, process } = arguments;
```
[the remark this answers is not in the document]
dgostencnik: Was added in both, `apply` and `reduce_dimension`, but it is probably missing also in some other processes. I don't think context is  used in any predefined process.

mitjak (approved):

PR #61, opened 2022-03-25
Filter temporal process

mitjak:
https://git.sinergise.com/team-6/openeo-platform/-/issues/129

### DIFF
--- a/src/pg_to_evalscript/javascript_datacube/DataCube.js
+++ b/src/pg_to_evalscript/javascript_datacube/DataCube.js
@@ -92,7 +92,7 @@ class DataCube {
                 throw new Error("Invalid ISO date string in temporal dimension label.");
             }
 
-            if (date.value >= start.value && date.value < end.value) {
+            if ((start === null || date.value >= start.value) && (end === null || date.value < end.value)) {
                 indices.push(i);
             }
         }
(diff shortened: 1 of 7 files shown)


src/pg_to_evalscript/javascript_datacube/DataCube.js
```diff
@@ -78,6 +109,50 @@ class DataCube {
             );
     }
 
+    filterTemporal(extent, dim) {
+        const dimensionName = dim ? dim : this.temporal_dimension_name;
```
Ardweaden, line 113: > If no specific dimension is specified or it is set to null, the filter applies to all temporal dimensions

So there could be multiple dimensions of type `TEMPORAL`, or there could be one but not the default one (that goes by the name `temporal_dimension_name`).

Ardweaden (changes requested):

src/pg_to_evalscript/javascript_datacube/DataCube.js
mitjak: > So there could be multiple dimensions of type TEMPORAL, or there could be one but not the default one (that goes by the name temporal_dimension_name).

We handle the second case, but not the first one. I have somewhat a hard time imagining how this would work in practice. But technically I see no problem of iterating over all temporal dimensions and applying the same algorithm to all dimensions.

Ardweaden: > We handle the second case

I don't think we do, unless I'm missing something. 

> I have somewhat a hard time imagining how this would work in practice

Yes, it's definitely not a very common scenario, but we should still follow specs and this should be achieved pretty simply, as you mentioned. I think the most likely scenario is that the default temporal dimension gets renamed and `filter_temporal` is called with `dim=null`.

mitjak: Ok, we weren't entirely clear.

> I think the most likely scenario is that the default temporal dimension gets renamed and filter_temporal is called with dim=null.

We definitely need to address this case, I missed this one, will update. (Currently `dim` would have to be set accordingly for this to work).

> I have somewhat a hard time imagining how this would work in practice. But technically I see no problem of iterating over all temporal dimensions and applying the same algorithm to all dimensions.

This was referring to having multiple dimensions. Where we would then iterate over all of them. Is this relevant?

Ardweaden: Funny enough, there was even an issue to clarify this by Anže:

https://github.com/Open-EO/openeo-processes/issues/203

mitjak: Cool, this issue confirms clearly that it should indeed be applied to all temporal dimensions 👍

Ardweaden: Looks good.

Ardweaden (approved):

PR #63, opened 2022-03-28
mask process

zcernigoj:
From the [docs](https://processes.openeo.org/#mask)

> The data cubes have to be compatible so that each dimension in the mask must also be available in the raster data cube with the same name, type, reference system, resolution and labels. Dimensions can be missing in the mask with the result that the mask is applied for each label of the missing dimension in the data cube. The process fails if there's an incompatibility found between the raster data cube and the mask.

Check is performed for **dimensions'** `name`, `type`, `labels`.
Not sure how to check for `reference system` and `resolution`.

Created an [issue](https://github.com/Open-EO/openeo-processes/issues/359) for clarification of the second sentence. -> Solved by implementing and using a function for broadcasting the `ndarray`.

---

### Tests:

**successful cases**

- different `mask` values: 
  - `number` [DONE]
  - `boolean` [DONE]
- different `replacement` values:
  - `number` [DONE]
  - `boolean` [DONE]
  - `string` [DONE]
  - `null` [DONE]
- dimensions missing in `mask` parameter
  - `bands` dimension missing [DONE]
  - `temporal` dimension missing [DONE]
  - `spatial` dimension missing
    - test with dimension of type `other` [DONE]

**tests for error handling**

- `replacement` wrong type [DONE] 
- dimensions missing in `data` parameter [DONE]
- different dimensions properties
  - different type [DONE]
  - different labels [DONE]
  - different reference system [not sure how to check]
  - different resolution [not sure how to check]


tests/unit_tests/test_mask.py
```diff
@@ -0,0 +1,374 @@
...
+                "data": data_3bands_3dates,
+                "mask": [mask_3bands_3dates_num[0]],
+                "scenes_data": scenes_3dates,
+                "scenes_mask": None,
+            },
+            "maskCube.removeDimension('temporal_name');",
+            resultWithReplacement(result_mask_no_temporal),
+        ),
+        # (  # test mask parameter missing dimensions: mask parameter with no bands dimension
+        #     # doesn't work properly: same result as for "mask parameter with no temporal dimension"
```
Ardweaden, line 178: We should "help" the broadcasting method by restricting the options how the smaller array can be broadcasted in the bigger one.

`[0,11,0]` could be validly broadcasted into an array of shape `(3,3)` either row-wise
```js
[
  [0, 11, 0],
  [0, 11, 0],
  [0, 11, 0]
]
```
or column-wise
```js
[
  [ 0,  0,  0],
  [11, 11, 11],
  [ 0,  0,  0]
]
```
and the method can't know we want the second one in this case.

We can help it by inserting the missing dimensions into the `mask` as a label-less dimensions of length 1. So in our case  `[0,11,0]` would become `[[0],[11],[0]]` or `ndarray([0,11,0], [3,1])`. Now we only one option for broadcasting.

src/pg_to_evalscript/javascript_processes/mask.js
```diff
@@ -0,0 +1,60 @@
...
+
+  let newData = data.clone();
+  let newDataFlat = newData.flattenToArray();
+
+  const maskBroadcasted = broadcastNdarray(mask.data, newData.data.shape);
+  const maskBroadcastedFlat = flattenToNativeArray(maskBroadcasted);
+
+  const replacement_val = replacement === undefined ? null : replacement;
+
+  for (let i = 0; i < newDataFlat.length; i++) {
```
Ardweaden, line 50: Of course the approach with flattening the arrays and comparing values will only work if the dimensions will be in the same order in the `DataCube`. If the dimensions would be switched, the values that are being masked here have different labels than the mask value. 

I would say that in the majority of the cases the dimension order will be the same, but this should either work with dimensions in arbitrary order or raise and error of order doesn't match.

Ardweaden (changes requested):

tests/unit_tests/test_mask.py
zcernigoj: Thanks, I added inserting of the missing dimensions in the `mask` argument as label-less dimensions of length 1.

I also added ordering the dimensions in the `mask` argument so that they are in the same order as in the `data` argument. This then guarantees that positions of the elements in the flattened arrays of both arguments are synced in terms of dimensions. (the solution for the [comment below](https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/63#discussion_r869269031))

src/pg_to_evalscript/javascript_processes/mask.js
zcernigoj: I realized that this problem and the problem mentioned above are connected, so I replied in the comment above.

tests/unit_tests/test_mask.py
Ardweaden: Wonderful, can we please add a test for this?

tests/unit_tests/test_mask.py
```diff
@@ -0,0 +1,371 @@
...
+            vars_definitions + f"const replacement = {replacement_parameter};"
+        )
+
+    additional_js_code_to_run = (
+        load_datacube_code()
+        + vars_definitions
+        + (additional_js_code_specific_to_case or "")
+    )
+
+    arguments = f"'data': dataCube, 'mask': maskCube, 'scenes': {scenes_data}"
```
Ardweaden, line 232: I think `scenes` does not need be passed to `mask` process.

tests/unit_tests/test_mask.py
zcernigoj: I added the test and while making it, I realized that there was a bug in `flattenToNativeArray()` function.

I removed the check for the subarray (code below) because with this check, it didn't use the `shape` and `stride` which are important for the transposed ndarrays - when the ndarray is transposed (`ndarray.transpose()`), only the `shape` and `stride` are changed, the `data` of the ndarray stays the same - and thus returned the wrong result.

```
if (isNotSubarray(ndarray, shape)) {
  return ndarray.data
}
```

This might affect the behaviour of the other processes that use `flattenToNativeArray()` / `DataCube.flattenToArray()`, but the tests are all successful, so it doesn't seem this result in a wrong things to be returned for other processes.

tests/unit_tests/test_mask.py
zcernigoj: True, thanks. Removed.

tests/unit_tests/test_mask.py
Ardweaden: Well, we never had a situation like that so `flattenToNativeArray` wasn't designed to handle it. However, simply returning the array is surely much faster that iterating over it. So I'd prefer if we added a flag (`transposed=true` or whatever), so that the simple approach would still be used, as the majority of the cases should use it. 

Of course we should avoid ever returning a transposed array from a process, but that's already the case now as other processes do not expect it.

Of course we can add `.slice()` for consistency.

-----------------------------------

Other than that, are you sure the tests work correctly? I find them a bit hard to read (I prefer the approach of e.g. `test_merge_cube`, i.e. passing data, shape and dimensions and setting them directly, but it's not a big deal) so I'm not sure if I'm misunderstanding them or not. 

Given it's being transposed and the dimensions switched after initialisation, I assume `maskCube` data should be the same as in the previous test (instead of that `mask_3bands_3dates_num[0],`). Because it's essentially transposed twice then.

Ardweaden (approved):

tests/unit_tests/test_mask.py
zcernigoj: > Ok I think they seem ok, It's setup as every second band to be masked, then transposed to every second temporal element.

Yes, it was like that.
I agree that it was hard to read. I changed the approach to the same as in `test_merge_cube.py`.

Ardweaden: Looks good.

Ardweaden (approved):

PR #64, opened 2022-03-31
Fix issue with spread operator for objects

Ardweaden:
This is not supported by V8 version `7.4` that SH is using and the evalscripts fail. 

I guess we'll have to run the evalscripts in V8 instead of node at some point. I tried nodejs14 which supposedly uses V8 7.4 but it passed.

### DIFF
--- a/src/pg_to_evalscript/javascript_processes/merge_cubes.js
+++ b/src/pg_to_evalscript/javascript_processes/merge_cubes.js
@@ -11,7 +11,7 @@ function merge_cubes(arguments) {
   }
 
   if (overlap_resolver && context) {
-    overlap_resolver.context = { ...context, ...overlap_resolver.context };
+    overlap_resolver.context = Object.assign(context, overlap_resolver.context);
   }
 
   for (let dimension of cube1.dimensions) {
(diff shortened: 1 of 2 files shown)


mitjak:
Fix looks ok. There are other occurrences of spread operator used on objects, is that not problematic?

Ardweaden:
> Fix looks ok. There are other occurrences of spread operator used on objects, is that not problematic?

Thanks, I checked again and I indeed found one more in `merge_cubes`, now there shouldn't be any more hopefully.

mitjak:
I believe elsewhere spread is only used on arrays, so we should be good. 

Just a note that `Object.assign` modifies original object. It might be safer to use `Object.assign({}, ..., ...);`

I don't think it's an issue with modifying `arguments` though.

Ardweaden:
> I believe elsewhere spread is only used on arrays, so we should be good.
> 
> Just a note that `Object.assign` modifies original object. It might be safer to use `Object.assign({}, ..., ...);`
> 
> I don't think it's an issue with modifying `arguments` though.

@mitjak You're right, fixed.

mitjak (approved):

PR #65, opened 2022-04-05
Aggregate temporal process

mitjak:
https://git.sinergise.com/team-6/openeo-platform/-/issues/142

This PR handles intervals in "date-time" format only. We will handle intervals in "date" and "time" format (see examples in docs) separately.

### DIFF
--- a/tests/unit_tests/test_aggregate_temporal.py
+++ b/tests/unit_tests/test_aggregate_temporal.py
@@ -115,7 +115,7 @@ def scenes():
 def test_aggregate_temporal(aggregate_temporal_process_code, data, scenes, example_input, expected_output):
     additional_js_code_to_run = (
         load_datacube_code()
-        + f"const cube = new DataCube({data}, 'bands_name', 'temporal_name', true, {scenes});"
+        + f"const cube = new DataCube({data}, 'bands_name', 'temporal_name', true, [], {scenes});"
     )
     process_arguments = f"{{...{json.dumps(example_input)}, 'data': cube, 'scenes': {scenes}, 'reducer': {example_input['reducer']}}}"
     output = run_process(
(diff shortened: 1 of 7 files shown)


dgostencnik (commented):
I think integration test(s) could also be added here, at least for [Example #1](https://processes.openeo.org/#aggregate_temporal)

dgostencnik (approved):

tests/unit_tests/test_aggregate_temporal.py
```diff
@@ -0,0 +1,235 @@
...
+)
+def test_aggregate_temporal_exceptions(aggregate_temporal_process_code, data, scenes, example_input, raises_exception, error_message):
+    additional_js_code_to_run = (
+        load_datacube_code()
+        + f"const cube = new DataCube({data}, 'bands_name', 'temporal_name', true, [], {scenes});"
+    )
+    process_arguments = f"{{...{json.dumps(example_input)}, 'data': cube, 'scenes': {scenes}, 'reducer': {example_input['reducer']}}}"
+    if raises_exception:
+        with pytest.raises(Exception) as exc:
+            run_process(
```
Ardweaden, line 223: Can `test_aggregate_temporal_exceptions` be changed to use `run_input_validation` method?

It turns out there is a major bug with the tests for exceptions discovered in https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/68. When that is merged, this test will fail on Windows devices.

The problem is that the `exc.value` (that we check includes `error_message`) is of form: `Command "node -e <entire-code> failed with exit code 1`.  So for as long your `error_message` string is in the source code, the test will pass...

`run_input_validation` correctly checks against `stderr` attribute of `subprocess.CalledProcessError`, but it also simplifies things.

Ardweaden (changes requested):

Ardweaden (approved):

PR #68, opened 2022-04-12
Fix issues with running tests on Windows

Ardweaden:
In Windows, a process command line is limited to 32 766 characters. So tests were failing because the evalscripts were too long. 
With this PR, code will be saved to a file on Windows and the command will just run the file.

However, some tests kept failing, and it uncovered a major bug. Some of the tests that expected an exception to happen used to pass despite being wrong and having all sorts of exception. That's because there were in form:
```
assert "some error string" in error.value
```
but the `error.value` looked like this:
```
command "node - e <entire code>" failed with exit code <exit code>
``` 
That meant that even though the process failed for a completely unrelated reason, for as long as the string we were checking against was in the source code, the test passed. 
However with
```
command "node <filename>" failed with exit code <exit code>
```
these tests failed.

The correct approach is using the `run_input_validation` method, which checks for the string in `stderr` attribute of `CalledProcessError`.

Also ran linting in the second commit, it was annoying before that some files were not linted, so I decided to just do it here.


tests/unit_tests/test_array_element.py
```diff
@@ -85,10 +85,18 @@ def test_array_element(array_element_process_code, example_input, expected_outpu
...
-    else:
-        run_process(array_element_process_code, "array_element", example_input)
+    additional_js_code_to_run = (
+        load_datacube_code()
+        + f"const d = {json.dumps(example_input['data']) if 'data' in example_input else 'undefined'};"
+        + f"d.labels = {json.dumps(example_input['labels']) if 'labels' in example_input else 'undefined'};"
+        if "data" in example_input
+        else ""
+    )
+    process_arguments = f"{{...{json.dumps(example_input)}, 'data': {'d' if 'data' in example_input else 'undefined'}}}"
```
zansinergise, line 95: I believe that this `'d'` will be interpreted as a string and not as a variable `d` when JS code will be run. I am not actually 100% sure about this, if I am wrong then fftr, just wanted to point this out and make sure everything is ok.

tests/unit_tests/test_array_labels.py
```diff
@@ -36,30 +36,24 @@ def test_array_labels(array_labels_process_code, example_input, expected_output)
...
 )
 def test_array_labels_exceptions(array_labels_process_code, example_input, raises_exception, error_message):
     additional_js_code_to_run = (
-        f"const d = {json.dumps(example_input['data'] if 'data' in example_input else 'undefined')};"
+        f"const d = {json.dumps(example_input['data']) if 'data' in example_input else 'undefined'};"
         + f"d.labels = {json.dumps(example_input['labels']) if 'labels' in example_input else 'undefined'};"
+        if "data" in example_input
+        else ""
+    )
+    process_arguments = f"{{...{json.dumps(example_input)}, 'data': {'d' if 'data' in example_input else 'undefined'}}}"
```
zansinergise, line 52: Same issue as above. Maybe it would be safer to do this 2 lines up (line 50) in the else statement and just write there `const d = undefined;` (as I am not sure if both `'d'` and `'undefined'` will be processed as strings or variables) and then in this line just pass `d` like `{{... , 'data': d}}`

zansinergise (changes requested):

tests/unit_tests/test_array_element.py
Ardweaden: Well, one can always test it out. For example, we can put the following in the python console:
```python
example_input = {"data": 1}
print(f"{{...{json.dumps(example_input)}, 'data': {'d' if 'data' in example_input else 'undefined'}}}")
>> {...{"data": 1}, 'data': d}
```
There are no quotes around `d`, so it will be interpreted as a variable. Let's check that in JS console:
```js
{...{"data": 1}, 'data': d}
>> Uncaught ReferenceError: d is not defined
```
Similar for undefined:
```python
example_input = {}
print(f"{{...{json.dumps(example_input)}, 'data': {'d' if 'data' in example_input else 'undefined'}}}")
>> {...{}, 'data': undefined}
```
In JS:
```js
{...{}, 'data': undefined}
>> {data: undefined}
```
-----------------------------
There used to be a bug in line 47 of `test_array_labels` though. I changed the line:
```python
 f"const d = {json.dumps(example_input['data'] if 'data' in example_input else 'undefined')};"
 ```
to
```python
f"const d = {json.dumps(example_input['data']) if 'data' in example_input else 'undefined'};"
```
You can see the parenthesis were placed incorrectly. If there is no `data` in `example_input`, the first expression can be simplified and evaluates to 
```python
f"const d = {json.dumps('undefined')};"
>> const d = "undefined"
```
and the second one to 
```python
f"const d = {'undefined'};"
>> const d = undefined
```
In the first case, undefined is a string and thus won't be the actual JS `undefined` object. In the second case it will be instantiated correctly.
----------------------------------
>  just wanted to point this out and make sure everything is ok

It's great to do this when one isn't certain about something. It's better to ask than regret not saying it out loud when things eventually fail. 
However, if there is a simple way of testing the problem, it's also good to take a few minutes to investigate it.

zansinergise: I see now yes, thanks for the explanation.

I agree I should check it, as soon as I started reading your comment I asked myself "why didn't I just do that already by myself?".

Resolving

zansinergise (approved):
Looks good 👍

PR #69, opened 2022-04-15
Use run_input_validation function in all unit tests

zansinergise:
First commit:
- remove all existing occurrences of "old" test blocks for input validation and instead use `run_input_validation` function

Second commit:
- run linting for a few files which were not linted


zansinergise:
> This looks ok. We might consider adding lint check to the pre-commit hook. Manually running it is tedious and it makes it hard to CR.

Yes I agree, one can easily forget to run linting which can be annyong

PR #70, opened 2022-04-25
Fix extendFinalDimensionWithData function

zansinergise:
Closes https://git.sinergise.com/team-6/openeo-platform/-/issues/152


src/pg_to_evalscript/javascript_datacube/DataCube.js
```diff
@@ -269,20 +269,36 @@ class DataCube {
...
+        }
+        const finalData = this.insertIntoFinalDimension(dataToAdd, loc, dataToAdd.length / elementsToInsertAtOnce, axis);     
         finalShape[axis]++;
-        
-        const finalData = this.insertIntoFinalDimension(dataToAdd, finalShape[axis], finalShape[axis] - 1)     
         this.data = ndarray(finalData, finalShape)
     }
 
-    insertIntoFinalDimension(dataToAdd, dimensionSize, locationInDimension) {
+    insertIntoFinalDimension(dataToAdd, offset, elementsToInsertAtOnce, axis) {
```
Ardweaden, line 286: This method should be renamed as it not longer works only for the final dimension.

Ardweaden, line 286: I don't think the `offset` works correctly, e.g.
```js
const d = new DataCube(ndarray([1,2,3,4,5,6,7,8],[2,2,2]), 'b','t', false)
d.insertIntoFinalDimension([9,99,999,9999],1,1,2) // insert into last dimension with offset 1
>>> [ 1, 9, 2, 99, 3, 999, 4, 9999,  5, 6, 7, 8]
```
Whereas I would expect `[ 1, 9, 2, 3, 99, 4, 5, 999, 6, 7, 9999, 8]`.
The method can perhaps be changed to just extend a dimension, not necessarily insert with an offset.

I missed you were working on this, so a general method that should be able to insert with offset was added in this PR: https://github.com/openEOPlatform/openeo-pg-evalscript-converter/pull/71.

Ardweaden (changes requested):

src/pg_to_evalscript/javascript_datacube/DataCube.js
zansinergise: Hm I guess this does not work correctly yes, I did not notice this before because I always used the `extendFinalDimensionWithData` function which calculates all the parameters by itself automatically.

Should we then just discard these changes I made and use the function that you wrote which works correctly?

zansinergise: Should be fixed

PR #72, opened 2022-05-03
Aggregate temporal date and time format

mitjak:
https://git.sinergise.com/team-6/openeo-platform/-/issues/142


zansinergise (approved):
Looks good to me 👍

src/pg_to_evalscript/javascript_common/common.js
```diff
@@ -45,6 +45,46 @@ function parse_rfc3339(dt, default_h = 0, default_m = 0, default_s = 0) {
...
+      }
+
+      return {
+        type: "time",
+        value: new Date(date).toISOString(),
+      };
+    }
+  } catch (err) {}
+
+  return null;
```
mkadunc, line 85: This seems like a lot of code for something that could be implemented with `Date.parse("1900-01-01T"+t)` (with potential checking of the regex and returning null if the string doesn't match)... am I missing something?

mitjak: Was thinking about the same thing actually. @dgostencnik is there any specific reason we are parsing datetime strings manually (function above) or could this be streamlined?

dgostencnik: I imagine there is a better way to do this. I think there was a problem with some edge cases for compare processes.

mitjak: I would guess problems could arise because `Date.parse()` is quite liberal. If we start with a (stricter) regex and only pass "approved" strings to the parsing function it might behave as expected.

mkadunc: I suggest we try:
- replacing all custom ISO-date parsing code with the `Date()` constructor and see if tests pass (leave the old implementations around, e.g. rename the function to `parse_rfc3339_ownimpl` just in case we need to revert, and write a new implementation of `parse_rfc3339`)
- the edge cases mentioned by Damjan should show up as failures
- if the tests are for processes rather than common functions, we should write new tests that isolate the problem to the time-parsing functions
- if it turns out that we actually need custom parsing, let's add a comment to the code explaining why this was needed

mitjak: Agree. I'll go ahead with merging this branch and update date parsing in a separate pull request.

PR #73, opened 2022-05-12
Fix/update date parsing

mitjak:
https://git.sinergise.com/team-6/openeo-platform/-/issues/157


src/pg_to_evalscript/javascript_common/common.js
```diff
@@ -3,81 +3,37 @@ function parse_rfc3339(dt, default_h = 0, default_m = 0, default_s = 0) {
...
-      if (g[6] !== undefined && g[6] !== "Z") {
-        // offset in minutes
-        const offset =
-          (parseInt(g[9] || 0) * 60 + parseInt(g[10] || 0)) *
-          (g[8] === "+" ? -1 : 1);
-        // add offset in miliseconds
-        date = date + offset * 60 * 1000;
-      }
+      const split = t.split(new RegExp('[Tt]'));
+      const time = split.length === 1 ? split[0] : split[1];
```
mkadunc, line 32: Do we actually need to support parsing of just the time part of date-time? From the function name I would expect it to fail on strings that contain the date part.

mitjak: Parsing time from date-time strings is used for time based comparisons. With all datetimes being normalised to the same date it's simple to make time-only comparisons. 

Example: 
Input date-time:
```
"2022-03-16T03:30:00.000Z"
"2022-03-17T03:00:00.000Z"
"2022-03-20T04:30:00.000Z"
```

Interval to compare by:
```
["03:00:00Z", "04:00:00Z"]
```

mkadunc: I see - I agree that such a function is useful; but I wonder whether it should really be called `parse_rfc3339_time`.. Vanilla implementation of what I intuitively expect when reading `parse_rfc3339_time` would assume that input is compatible with `full-time` or `partial-time` from [the RFC](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6); and would fail otherwise... The two use-cases we have for this[^1] are IMO distinct enough that we wouldn't want to mix them up in the same function.

Or we should at least make it clear from the name of the function that it doesn't really parse its input, but rather extracts information from a subset of its input.

[^1]: The two use-cases being a) I have a string that I assume is time-of-day and would like to convert it to a data object that I can work with vs. b) I have a string that I know is date-time and I would like to strip the date information to get a time-of-day data object

mitjak: I agree the naming is not nearly descriptive enough. But thinking about a better name I think also the purpose for such function is not intuitive at all.

What if instead of returning a date object normalised to `1900-01-01` we would rather return a number of seconds of the day. And name the function `extract_seconds_of_day()` or similar.

```
new Date('1970-01-01T' + matchTime[0]).getTime() / 1000
```

With this update I believe it would also make more sense to keep the two use cases together in the same function. As it would make sense to do so from the usage perspective.

mkadunc: I agree (assuming that this is what we need), but maybe go with milliseconds, so that we can return an `uint` rather than floating-point number; and we retain precision of the input

mitjak: > assuming that this is what we need

I believe it should be ok. If we ever need to format extracted time, we can still construct new date object from milliseconds.

tests/unit_tests/test_common.py
```diff
@@ -41,20 +41,20 @@ def test_common(common_code, example_input, expected_output):
...
-        ("12:34:56.123Z", dict({"type": "time", "value": "1900-01-01T12:34:56.123Z"})),
-        ("13:00:00+01:00", dict({"type": "time", "value": "1900-01-01T12:00:00.000Z"})),
-        ("11:00:00-01:30", dict({"type": "time", "value": "1900-01-01T12:30:00.000Z"})),
-        ("00:00:00+01:00", dict({"type": "time", "value": "1899-12-31T23:00:00.000Z"})),
-        ("2018-01-01T12:00:00Z", dict({"type": "time", "value": "1900-01-01T12:00:00.000Z"})),
-        ("2018-01-01T12:34:56.123Z", dict({"type": "time", "value": "1900-01-01T12:34:56.123Z"})),
-        ("2018-01-01T13:00:00+01:00", dict({"type": "time", "value": "1900-01-01T12:00:00.000Z"})),
-        ("2018-01-01T11:00:00-01:30", dict({"type": "time", "value": "1900-01-01T12:30:00.000Z"})),
-        ("2018-01-01T00:00:00+01:00", dict({"type": "time", "value": "1899-12-31T23:00:00.000Z"})),
+        ("17:35:58Z", dict({"type": "milliseconds-of-day", "value": 63358000})),
```
Ardweaden, line 44: Minor: no need to use the `dict()` constructor as you pass is already a dictionary.

src/pg_to_evalscript/javascript_common/common.js
```diff
@@ -1,83 +1,33 @@
...
-        date = date + offset * 60 * 1000;
-      }
+    const matchDateTime = dateTime.match(regexDateTime);
+    const matchDate = dateTime.match(regexDate);
 
+    if (matchDateTime || matchDate) {
       return {
-        type: dt.match(regexDate) ? "date" : "date-time",
-        value: new Date(date).toISOString(),
+        type: matchDate ? "date" : "date-time",
```
Ardweaden, line 12: Can we put all these types (including `"milliseconds-of-day"`) into constants and use that, to avoid issues with typos.

Ardweaden (changes requested):

Ardweaden (approved):

PR #81, opened 2022-06-10
Use older version of Node to run tests and remove object literals spr…

Ardweaden:
Closes https://git.sinergise.com/team-6/openeo-platform/-/issues/171

Tried using V8 directly instead, but it's much simpler to just use an older version of NodeJS. Object literal spreading was introduced in `8.3.0`, so we need to use `8.2.1`.

Unfortunartely there is no `python-nodejs` images for that combination.

Removes all the spread operators where necessary, also older version order identical elements in the opposite way (`order` process). You can check, however, that the current expected outputs in `test_order` match the output produced by evalscript on SH.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,11 @@
-FROM nikolaik/python-nodejs:python3.7-nodejs6
+FROM python:3.8.13
 
 ADD . .
+
+SHELL ["/bin/bash", "--login", "-c"]
+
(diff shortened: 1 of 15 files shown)


mitjak (approved):
Looks good.

The issue with sort stability in `sort` is described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability

PR #91, opened 2023-11-09
Feature - data fusion support

zansinergise:
https://github.com/Open-EO/openeo-sentinelhub-python-driver/issues/24


src/pg_to_evalscript/evalscript.py
```diff
@@ -103,10 +103,24 @@ def write_ndarray_definition(self):
...
+
+        return datacube_creation
 
     def write_runtime_global_constants(self):
-        return f"const INPUT_BANDS = {self.input_bands};"
+        all_bands = []
+        for datasource_with_bands in self.input_bands:
+            all_bands.extend(datasource_with_bands["bands"])
+
+        return f"const INPUT_BANDS = {list(set(all_bands))};"
```
zcernigoj, line 123: How does it work if there are 2 different collections with same band names (advanced: representing different wavelength, e.g. S2, L8)?

If I understand correctly, code generated in `makeArrayFromSamples` creates labels for `band` dimension for all datacubes with all unique band names.

Would this be a problem in some usecase?

src/pg_to_evalscript/evalscript.py
```diff
@@ -131,13 +151,21 @@ def write_update_output(self):
...
     def write_output_variable(self):
         if len(self.nodes) == 0:
-            return self.initial_data_name
+            return "_".join(self.initial_data_names.values())
         return self.nodes[-1].node_varname_prefix + self.nodes[-1].node_id
 
     def determine_output_dimensions(self):
         dimensions_of_inputs_per_node = defaultdict(list)
+        all_bands = []
+        for datasource_with_bands in self.input_bands:
```
zcernigoj, line 160: Can we have some comment somewhere how one element (`datasource_with_bands`) of `self.input_bands` looks like because even now, I'm constantly not sure.

tests/process_graphs/bands_null_graph.json
```diff
@@ -1,5 +1,5 @@
 {
-   "loadco1":{
+   "1":{
```
zcernigoj, line 2: Were the changes in the ids of these processes (here and test_filter_temporal.json, test_resample_cube_temporal_downsample.json, test_resample_cube_temporal_upsample.json) needed for a particular reason?

src/pg_to_evalscript/evalscript.py
zcernigoj: You can ignore the above comment, I got confused because of the way the `all_bands.extend()` is written.

I think it would be clearer if it would be like:

```python
        for datasource_with_bands in self.input_bands:
            if datasource_with_bands is not None and datasource_with_bands["bands"] is not None:
                all_bands.extend(datasource_with_bands["bands"])

```

zcernigoj (changes requested):
Looks pretty ok. I would make some changes and I'm not sure about the band names not being split by datasource. See comments for details

src/pg_to_evalscript/evalscript.py
zansinergise: Changed

src/pg_to_evalscript/evalscript.py
zansinergise: > How does it work if there are 2 different collections with same band names (advanced: representing different wavelength, e.g. S2, L8)?

This should not be a problem I believe, since when each DataCube is generated, each gets its own bands metadata, so even if they have the same band names, the wavelength in this case would differ for each collection/DataCube because of different bands metadata.

> If I understand correctly, code generated in makeArrayFromSamples creates labels for band dimension for datacubes for all datasources with all unique band names, even if some datasource doesn't have that band.

I am not 100% sure what would happen if some datasource doesn't have that band, I need to rethink and try this out - will get back to you with what I figure out

zcernigoj: > If I understand correctly, code generated in makeArrayFromSamples creates labels for band dimension for datacubes for all datasources with all unique band names, even if some datasource doesn't have that band.

I tested this with S2L1C (B10 not present in S2L2A) and S2L2A (SCL not present in S2L1C) and it worked and generated correct results, so I think this can be solved in a separate MR in the future.

src/pg_to_evalscript/evalscript.py
zcernigoj: Thanks

src/pg_to_evalscript/evalscript.py
```diff
@@ -120,8 +134,14 @@ def write_update_output(self):
...
-        collection_scenes_length = "* collection.scenes.length" * number_of_original_temporal_dimensions
-        number_of_final_dimensions = len(self._output_dimensions) + 1 if self.encode_result else 0
+        if len(self.input_bands) > 1:
+            collection_scenes_length = (
+                "* Object.values(collection).reduce((acc, val) => acc + val.scenes.length, 0)"
+                * number_of_original_temporal_dimensions
+            )
+        else:
+            collection_scenes_length = "* collection.scenes.length" * number_of_original_temporal_dimensions
+        number_of_final_dimensions = len(self._output_dimensions) + 1 if self.encode_result else 0  # 3
```
zcernigoj, line 144: ah, missed this comment - it can probably be deleted

tests/process_graphs/bands_null_graph.json
zcernigoj: Thanks, I understand now. Would it make sense to change them all to `loadcollection` instead of just `1`?

zcernigoj (changes requested):

src/pg_to_evalscript/evalscript.py
zcernigoj: issue for later https://github.com/openEOPlatform/openeo-pg-evalscript-converter/issues/93

tests/process_graphs/bands_null_graph.json
zansinergise: Changed them to `loadcollection` but this meant also changing a few other PGs

src/pg_to_evalscript/evalscript.py
zansinergise: removed

tests/process_graphs/bands_null_graph.json
zcernigoj: Thanks, looks good

zcernigoj (approved):

